PR #19, opened 2016-08-13
Notify the share owner when another user downloads a shared file

oparoz:
Part 2 of the solution for nextcloud/server#853
You will find part 1 at https://github.com/nextcloud/server/pull/854

@nickvergessen

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -112,6 +112,8 @@ class Application extends App {
 		Util::connectHook('\OCA\Files_Trashbin\Trashbin', 'post_restore', FilesHooksStatic::class, 'fileRestore');
 		Util::connectHook('OCP\Share', 'post_shared', FilesHooksStatic::class, 'share');
 		Util::connectHook('OCP\Share', 'pre_unshare', FilesHooksStatic::class, 'unShare');
+		Util::connectHook('\OC\Files\Storage\Shared', 'fopen', FilesHooksStatic::class, 'downloadedShare');
+		Util::connectHook('\OC\Files\Storage\Shared', 'file_get_contents', FilesHooksStatic::class, 'downloadedShare');
 	}
 
 	/**
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -1,8 +1,10 @@
 <?php
 /**
  * @copyright Copyright (c) 2016, ownCloud, Inc.
+ * @copyright Copyright (c) 2016, Olivier Paroz
  *
  * @author Joas Schilling <coding@schilljs.com>
+ * @author Olivier Paroz <developer@oparoz.com>
  *
  * @license AGPL-3.0
  *
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -1,10 +1,12 @@
 <?php
 /**
  * @copyright Copyright (c) 2016, ownCloud, Inc.
+ * @copyright Copyright (c) 2016, Olivier Paroz
  *
  * @author Frank Karlitschek <frank@karlitschek.de>
  * @author Joas Schilling <coding@schilljs.com>
  * @author Thomas Müller <thomas.mueller@tmit.eu>
+ * @author Olivier Paroz <developer@oparoz.com>
  *
  * @license AGPL-3.0
  *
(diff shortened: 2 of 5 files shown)


lib/FilesHooks.php
```diff
@@ -540,13 +572,23 @@ protected function addNotificationsForUser($user, $subject, $subjectParams, $fil
...
-		$app = $type === Files_Sharing::TYPE_SHARED ? 'files_sharing' : 'files';
+		$app = ($type === Files_Sharing::TYPE_SHARED || $type === Files_Sharing::TYPE_REMOTE_DOWNLOAD) ? 'files_sharing' : 'files';
 		$link = $this->urlGenerator->linkToRouteAbsolute('files.view.index', array(
 			'dir' => ($isFile) ? dirname($path) : $path,
 		));
 
 		$objectType = ($fileId) ? 'files' : '';
 
+		$client = "Web";
+		if ($this->request->isUserAgent([Request::USER_AGENT_OWNCLOUD_DESKTOP,])) {
```
nickvergessen: Use public API?
https://github.com/nextcloud/server/blob/ba87db3fccb40aa58d84d12932424c83a4db411f/lib/public/IRequest.php#L66-L79

oparoz: Fixed, thanks!

MorrisJobke:
Tested and works 👍

MorrisJobke:
@nickvergessen feel free to merge

blizzz:
Is it possible to have these similar messages grouped? 

spectacle je5710

e.g. "… was downloaded 7 times" and "was downloaded by Alice, Bob, Chloe via the Web client"

Or is it out of scope? 

This so far works also between LDAP users :+1:

oparoz:
Thanks for testing with LDAP users :)
I think it would be out of scope and something Activity and the activity tab would have to deal with. One problem would be that you lose the time stamp. In your screenshot about public shares, it's not a problem since users are anonymous, but with this PR, you get a notification per user and wouldn't want to mix them up.

nickvergessen:
We have grouping in place, I will test and review if it's possible and otherwise merge as is.

lib/AppInfo/Application.php
```diff
@@ -101,6 +101,8 @@ public function registerFilesActivity() {
 		Util::connectHook('\OCA\Files_Trashbin\Trashbin', 'post_restore', 'OCA\Activity\FilesHooksStatic', 'fileRestore');
 		Util::connectHook('OCP\Share', 'post_shared', 'OCA\Activity\FilesHooksStatic', 'share');
 		Util::connectHook('OCP\Share', 'pre_unshare', 'OCA\Activity\FilesHooksStatic', 'unShare');
+		Util::connectHook('\OC\Files\Storage\Shared', 'fopen', 'OCA\Activity\FilesHooksStatic', 'downloadedShare');
+		Util::connectHook('\OC\Files\Storage\Shared', 'file_get_contents', 'OCA\Activity\FilesHooksStatic', 'downloadedShare');
```
nickvergessen: @oparoz this only works for local shares, not remote shares. Is that intended? I would think logging remote downloads is more important, then local once?

oparoz: Nope. I'm guessing the missing hooks mean that this would not work with federated clouds.

nickvergessen: Yeah, so this is intended for federation mostly? Then I will have a look how we can do that.

oparoz: This is for both. I've only used it with local users, so didn't experience the missing events, but it would be good to make it work with remote shares as well. I had a quick look at the code and think the storage wrapper is where there should be something of interest for this, since this is what the webdav endpoint uses:
https://github.com/nextcloud/server/blob/638f852ed85af759a0ce71397a1f23080c5cf14e/apps/dav/appinfo/v1/publicwebdav.php

nickvergessen: Yeah this seems to be the only way I can see this working atm is indeed a storage wrapper. The problem is we don't have the file id available at that point so this is a bit taffer

nickvergessen:
Replaced with this new app, so the feature can be disabled by default (not installing the app)
https://github.com/nextcloud/files_downloadactivity

PR #22, opened 2016-08-16
Use the last known user instead of erroring

nickvergessen:
Similar to https://github.com/owncloud/activity/pull/531

### Steps
1. Create remote share from a@a to b@b
2. As b@b upload a file into the share
3. As b@b check the activity stream for a "You created..." entry :x:

Should backport to 10 and 9


nickvergessen:
10: https://github.com/nextcloud/activity/pull/23
9: https://github.com/nextcloud/activity/pull/24

rullzer:
:+1:

PR #25, opened 2016-08-17
Try to identify remote users by their token

nickvergessen:
- [x] Tests

### Steps
1. Create a federated share
2. As recipient accept the share
3. As recipient upload a file
4. As share owner open the activity feed

### Before
`"remote user" created a.txt`

### After
`test@... created a.txt`

PS: you need to do a new upload, since the value is stored in the DB on upload (it's not only reendering)

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -41,6 +41,9 @@ use OCA\Activity\ViewInfoCache;
 use OCP\AppFramework\App;
 use OCP\AppFramework\IAppContainer;
 use OCP\IContainer;
+use OCP\IUser;
+use OCP\Share;
+use OCP\Share\Exceptions\ShareNotFound;
 use OCP\Util;
 
 class Application extends App {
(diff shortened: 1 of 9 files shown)


lib/AppInfo/Application.php
```diff
@@ -176,7 +179,24 @@ public function __construct (array $urlParams = array()) {
...
+			if ($user instanceof IUser) {
+				return (string) $user->getUID();
+			}
+
+			$request = $server->getRequest();
+			if (!empty($request->server['PHP_AUTH_USER'])) {
+				$token = $request->server['PHP_AUTH_USER'];
+				try {
+					$share = $server->getShareManager()->getShareByToken($token);
+					if ($share->getShareType() === Share::SHARE_TYPE_REMOTE) {
```
nickvergessen: @rullzer is that the correct thing to do? Because the constant is marked with:
```
// ToDo Check if it is still in use otherwise remove it
```

lib/AppInfo/Application.php
```diff
@@ -176,7 +179,24 @@ public function __construct (array $urlParams = array()) {
...
 
 			$user = $server->getUserSession()->getUser();
-			return ($user) ? $user->getUID() : '';
+			if ($user instanceof IUser) {
+				return (string) $user->getUID();
+			}
+
+			$request = $server->getRequest();
+			if (!empty($request->server['PHP_AUTH_USER'])) {
+				$token = $request->server['PHP_AUTH_USER'];
```
nickvergessen: @LukasReschke is there a cleaner way to access this information?

lib/AppInfo/Application.php
rullzer: Yep that is good. We use those constants all over the place. I might move them to the share manager at some point. But we need a deprecation period anyways.

lib/AppInfo/Application.php
LukasReschke: Not that I know of :see_no_evil:

nickvergessen:
@rullzer would be nice if you can review this.

rullzer:
Failing tests.
But looks good and a quick test shows it working.

rullzer:
Btw that cloudIdFormatter does it also look in the federated cloud user exchange thingy?

nickvergessen:
> Btw that cloudIdFormatter does it also look in the federated cloud user exchange thingy?

I ask the contacts manager, not sure if that provides them as well

nickvergessen:
Writing new tests doesn't fix existing tests :speak_no_evil: 

Adjusted, want to review again?

rullzer:
:+1:

PR #33, opened 2016-08-26
Use automatic dependency injection

nickvergessen:
@rullzer want to review some clean up?

Only thing left is the worst class of this app `OCA\Activity\FilesHooks` which needs some further examination.


rullzer:
LGTM! :tada:

PR #37, opened 2016-08-31
Move the APIv1 to OCSController

nickvergessen:
@rullzer


appinfo/routes.php
```diff
@@ -1,55 +1,36 @@
...
-	['name' => 'Activities#showList', 'url' => '/', 'verb' => 'GET'],
-	['name' => 'Feed#show', 'url' => '/rss.php', 'verb' => 'GET'],
-	['name' => 'EndPoint#getDefault', 'url' => '/api/v2/activity', 'verb' => 'GET'],
-	['name' => 'EndPoint#getFilter', 'url' => '/api/v2/activity/{filter}', 'verb' => 'GET'],
-]]);
+return [
+	'ocs' => [
+		['root' => '/cloud', 'name' => 'APIv1#get', 'url' => '/activity', 'verb' => 'GET'],
+//		['root' => '/apps/activity', 'name' => 'APIv2#getDefault', 'url' => '/api/v2/activity', 'verb' => 'GET'],
+//		['root' => '/apps/activity', 'name' => 'APIv2#getFilter', 'url' => '/api/v2/activity/{filter}', 'verb' => 'GET'],
```
nickvergessen: @rullzer is there a default `root` so I could skip this?
And what should be part of root and what part of the URL?

rullzer: Default root is `/apps/<app>` so in this case `/apps/activity`.
Basically it is a hack to allow the stuff of `/cloud/activity` etc.

I did not hard code any limitations so far. So

* `'root'='', path='/foo/bar'`
* `'root'='/foo', path='/bar'`
* `'root'='/foo/bar', path=''`

All map to the same. But so far I think the only place we need it is to do `'root'='/cloud', 'path'='/baz'`.

rullzer:
Awesome stuff! :+1:

PR #46, opened 2016-09-22
Add clipboard support to the RSS feed link

nickvergessen:
activity_-_nextcloud_-_2016-09-22_18 21 26

@jancborchardt want to review?

### DIFF
--- a/templates/stream.app.navigation.php
+++ b/templates/stream.app.navigation.php
@@ -45,7 +45,11 @@
 		<div id="app-settings-content">
 			<input type="checkbox"<?php if ($_['rssLink']): ?> checked="checked"<?php endif; ?> id="enable_rss" class="checkbox" />
 			<label for="enable_rss"><?php p($l->t('Enable RSS feed'));?></label>
-			<input id="rssurl"<?php if (!$_['rssLink']): ?> class="hidden"<?php endif; ?> type="text" readonly="readonly" value="<?php p($_['rssLink']); ?>" />
+
+			<span id="rssurl"<?php if (!$_['rssLink']): ?> class="hidden"<?php endif; ?>>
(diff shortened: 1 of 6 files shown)


templates/stream.app.navigation.php
```diff
@@ -45,7 +45,11 @@
 		<div id="app-settings-content">
 			<input type="checkbox"<?php if ($_['rssLink']): ?> checked="checked"<?php endif; ?> id="enable_rss" class="checkbox" />
 			<label for="enable_rss"><?php p($l->t('Enable RSS feed'));?></label>
-			<input id="rssurl"<?php if (!$_['rssLink']): ?> class="hidden"<?php endif; ?> type="text" readonly="readonly" value="<?php p($_['rssLink']); ?>" />
+
+			<span id="rssurl"<?php if (!$_['rssLink']): ?> class="hidden"<?php endif; ?>>
+				<input class="feed-link" type="text" readonly="readonly" value="<?php p($_['rssLink']); ?>" />
+				<a class="clipboardButton icon icon-clippy" data-clipboard-target="#rssurl input"></a>
```
jancborchardt: No need for the separate `icon` class here. :) Also, why use an additional class »clipboardButton« when you can simply use icon-clippy? And for classes please only use dashes, never camel case!

nickvergessen: This was all copy pasted ;)

But sure I can address this.

jancborchardt: That’s what I thought :D so we should fix it going forward that not more people copypaste useless code. ;)

nickvergessen: Updated

jancborchardt:
Seems to work :+1:

PR #48, opened 2016-10-03
add icons to navigation

jancborchardt:
Much nicer and easier to identify. Just like in Files and Mail. Before & after:
capture du 2016-10-03 18-24-07capture du 2016-10-03 18-23-41

Please review @nickvergessen @nextcloud/designers :)


raghunayyar:
👍

skjnldsv:
And svg are optimized! :+1: :open_mouth:

nickvergessen:
Let me have look, the icons should be added by the apps that provide the filters.

jancborchardt:
@nickvergessen the apps are all in core already (except the activity icon, which is in the app itself). But in files too for some reason we copied them over.

I think it was the naming of the entries. So if we can have an entry corresponding to the id (for example `$navigation['icon']`):

- all -> activity-dark
- self -> user
- by -> contacts-dark
- files_favorites -> star
- comments -> comment
- shares -> share

Then we can get rid of the extra classes and just use icon-star, icon-comment etc generated by the snipped `icon-<?php p($navigation['icon']) ?>` instead of being bound to the ID.

Also cc @schiessle @MorrisJobke so we could fix that in files as well and add an icon attribute.

nickvergessen:
I moved the icon thing to the API, so apps can specify their own icons. Core apps are handled in https://github.com/nextcloud/server/pull/1622

Please review that one as well.

Thanks @jancborchardt for the initial PR!

PR #50, opened 2016-10-05
make comment icon less intense

MorrisJobke:
cc @nickvergessen

Before:
bildschirmfoto von 2016-10-05 23-57-53

After:
bildschirmfoto von 2016-10-05 23-58-03

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -160,24 +160,27 @@
 	display: inline-block;
 	vertical-align: top;
 	margin-top: 2px;
-}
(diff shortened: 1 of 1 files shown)


nickvergessen:
I turned the filter around, so it reduces all images (apart from `.is-colored`) which I will add to the two icons that should not be less intense.
It's easier and more future proof this way.

MorrisJobke:
Looks good and works :+1:

PR #51, opened 2016-10-09
Clean all superposed white backgrounds

skjnldsv:
See https://github.com/nextcloud/server/pull/1657

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -46,7 +46,6 @@
 }
 
 #container {
-	background: #FFF;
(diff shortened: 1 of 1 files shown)


skjnldsv:
We can merge this one. Still a simple cleanup.
We don't need this background since we don't have superposition on the activity app and the background is white already!

@nextcloud/designers

MorrisJobke:
Makes sense and works 👍 

cc @nickvergessen

PR #52, opened 2016-10-12
Catch errors when one email could not be sent

nickvergessen:
Fix nextcloud/server#1716

### DIFF
--- a/lib/MailQueueHandler.php
+++ b/lib/MailQueueHandler.php
@@ -218,11 +218,12 @@ class MailQueueHandler {
 	 * @param string $lang Selected language of the recipient
 	 * @param string $timezone Selected timezone of the recipient
 	 * @param int $maxTime
+	 * @return bool True if the entries should be removed, false otherwise
 	 */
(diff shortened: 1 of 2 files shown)


nickvergessen:
Tested by @volker-raschek in https://github.com/nextcloud/server/issues/1716#issuecomment-253209087

Counting as 👍 

@MorrisJobke second review?

lib/MailQueueHandler.php
```diff
@@ -270,9 +271,15 @@ public function sendEmailToUser($userName, $email, $lang, $timezone, $maxTime) {
 		$message->setSubject((string) $l->t('Activity notification'));
 		$message->setPlainBody($emailText);
 		$message->setFrom([$this->getSenderData('email') => $this->getSenderData('name')]);
-		$this->mailer->send($message);
+
+		try {
+			$this->mailer->send($message);
+		} catch (\Exception $e) {
```
MorrisJobke: This is always really bad 🙈 Then an exception is simply dropped - at least having it in debug mode visible would make this traceable.

MorrisJobke:
:+1:

PR #55, opened 2016-10-19
design fixes for tags and avatars

jancborchardt:
Before
capture du 2016-10-19 10-54-39
After: nicer style for tags to be symmetric to how they look elsewhere, and some margin for the avatar:
capture du 2016-10-19 10-54-13

Please review @nickvergessen @nextcloud/designers

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -139,6 +139,7 @@
 	vertical-align: middle;
 	display: inline-block;
 	margin-right: 5px;
+	margin-left: 3px;
 }
 
 .activityTabView .activity.box {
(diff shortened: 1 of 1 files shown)


skjnldsv:
👍

eppfel:
It looks like the user names are not on the same baseline. Or do the avatars create this effect, because they should be positioned a bit higher?

jancborchardt:
@eppfel good catch, will investigate!

jancborchardt:
It seems to be the only setting where it’s clean for both avatars and avatar placeholders. I’d fix it separately because the rest works. :)

PR #72, opened 2016-11-15
Fix size of mimetype icons and spacing between previews

MorrisJobke:
cc @jancborchardt @nickvergessen @LukasReschke


MorrisJobke:
For https://github.com/nextcloud/server/issues/2134

PR #76, opened 2016-11-21
Clarify notify_self filter description

jowi24:
Removed "file" context from checkbox description, as this filter applies also to all actions contributed by other apps.

### DIFF
--- a/templates/personal.php
+++ b/templates/personal.php
@@ -63,7 +63,7 @@ style('activity', 'settings');
 	<br />
 	<input id="notify_setting_self" name="notify_setting_self" type="checkbox" class="checkbox"
 		value="1" <?php if ($_['notify_self']): ?> checked="checked"<?php endif; ?> />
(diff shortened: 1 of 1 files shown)


nickvergessen:
Actually that was changed for Nextcloud 10:
https://github.com/nextcloud/activity/commit/7a3adffec0489701584f54d20e88228131c6dd56
And it should only affect files. Which others do you see influenced by this?

jowi24:
I implemented an Activity extension in my own app using OCP\Activity\IExtension and recognized, that notification types provided via getNotificationTypes() and published via ActivityManager->publish are also affected by this setting. I'm fine with this behavior, because it reduces the number of notification types I have to provide in my own app, but I think the current label description is misleading.

nickvergessen:
Right, thanks.

Out of curiosity, which app is this about? ;) I tried to look at your profile but seems it's not there.

jowi24:
Thank you! I'm still in (early) development and have not yet uploaded the code. I'm currently experimenting with integration in the activity and notification apps and therefore came across this small issue.

PR #83, opened 2016-11-29
Fix the icon size of events to be max 16px

nickvergessen:
16px is intended and all icons so far used it. But the icon of the calendar is 32 pixels, therefor this adjustment is needed.

@MorrisJobke


LukasReschke:
LGTM

PR #93, opened 2016-12-14
shorten activity mail wording, fix * for bullets to -, time in parentheses

jancborchardt:
Stuff left to do I can’t figure out how to do:

- [ ] remove trailing slash in instance URL (currently `/cloud.example.com/`)
- [x] remove leading slash of any file (currently `/Some document.odt`) => https://github.com/nextcloud/server/pull/2695
- [x] remove seconds from timestamp (currently `today 18:45:00`)
- [ ] don’t show the bullet point `-` when there is only a single activity
- [ ] change title to `Nextcloud activity notification` for when multiple things happened and `Nextcloud: Joas shared Picture.jpg with you` for when it’s a single action. (of course with themed app name)

Can you help there @nickvergessen?

### DIFF
--- a/lib/MailQueueHandler.php
+++ b/lib/MailQueueHandler.php
@@ -251,7 +251,7 @@ class MailQueueHandler {
 
 			$relativeDateTime = $this->dateFormatter->formatDateTimeRelativeDay(
 				$activity['amq_timestamp'],
-				'long', 'medium',
+				'long', 'short',
 				new \DateTimeZone($timezone), $l
 			);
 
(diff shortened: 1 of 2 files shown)


nickvergessen:
Fixed the time stamp formatting

The leading slashes are dealt in https://github.com/nextcloud/server/pull/2695

I will not use 2 different emails for "single activity" and "multi activity"
Also changing the subject will break email filters that people use, so I suggest to not do that either.

=> all done, merging

jancborchardt:
Awesome, thanks a lot @nickvergessen! :)

PR #101, opened 2017-01-10
Use accesslist from the new share manager

### LINKED ISSUE #61: I'm receiving email notifications for shares i canceled.
### Steps to reproduce
1. Set up 2 accounts A and B.
2. From within A share a folder with B. Put some files in the folder.
3. If B enabled notifications in the settings, notifications will be send to B, if A adds files to the shared folder.
4. Now let B cancel the share.
5. The folder will disappear in account B.
### Expected behaviour

B should not receive email notifications, if A now adds files.
### Actual behaviour

B receives email notifications.
### Server configuration

**Operating system**:
Debian 8.6
**Web server:**
nginx 1.6.2
**Database:**
MySQL
**PHP version:**
5.6.27
**Nextcloud version:** (see Nextcloud admin page)
10.0.1 (stable)
**Where did you install Nextcloud from:**
https://nextcloud.com/install/
**Signing status:**
No errors have been found.

nickvergessen:
Ref https://github.com/nextcloud/server/pull/2834

Fix #61

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -26,10 +26,13 @@ namespace OCA\Activity;
 
 use OC\Files\Filesystem;
 use OC\Files\View;
+use OC\Share20\ShareHelper;
 use OCA\Activity\Extension\Files;
 use OCA\Activity\Extension\Files_Sharing;
 use OCP\Activity\IManager;
+use OCP\Files\IRootFolder;
 use OCP\Files\Mount\IMountPoint;
+use OCP\Files\Node;
 use OCP\Files\NotFoundException;
 use OCP\IDBConnection;
 use OCP\IGroup;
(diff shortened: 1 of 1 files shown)


lib/FilesHooks.php
```diff
@@ -499,7 +513,30 @@ protected function generateMoveActivities($users, $beforePathMap, $afterPathMap,
...
+			return [];
+		}
+
+		if (!$node instanceof Node) {
+			return [];
+		}
+		$accessList = $this->shareManager->getAccessList($node, true, true);
+		$helper = new ShareHelper($this->rootFolder); // FIXME
+
+		$nodes = $helper->getPathsForAccessList($node, $accessList['users']);
```
nickvergessen: This sets up all the users and is therefor very slow :/
Also I wonder about a potential memory issue with nodes being loaded for all users

rullzer: I don't think memory issues are the main concern here. Yes it is slow. But as discussed the old endpoint just returned false info in some cases.

Would it maybe make sense to just use the filename and link to the file by using not a path but the nodeid?

PR #124, opened 2017-03-21
Listen to the unshare event so we can generate activities for link ex…

nickvergessen:
…piration

Based on nextcloud/server#3844


rullzer:
test failing...

nickvergessen:
Fixed @rullzer mind reviewing again?

PR #126, opened 2017-03-25
Fix phpunit execution with php7

nickvergessen:
@MorrisJobke


MorrisJobke:
JS failures are most likely because of the change where sinon is imported in the server now :/ https://github.com/nextcloud/server/pull/4025

PR #129, opened 2017-03-27
fix position of previews, add a bit padding to the left

jancborchardt:
Before:
capture du 2017-03-28 00-50-11
After – nicely left-aligned with the text:
capture du 2017-03-28 00-49-13

Please review @nextcloud/designers @nickvergessen


nickvergessen:
Failing test is unrelated

PR #136, opened 2017-04-05
broadcast event on loading the Activities Controller

daita:
This would allow any app to add some stuff to the Activity app to add some interaction

Signed-off-by: Maxence Lange <maxence@nextcloud.com>


nickvergessen:
I fixed the test and also added an event object
I think it would be good, if you only do your stuff on `$event->getSubject() === 'all'`

PR #141, opened 2017-04-21
Use the new mail template

nickvergessen:
- [x] Requires: https://github.com/nextcloud/server/pull/4467 

@LukasReschke @MorrisJobke we are currently `htmlspecialchars` all the input for Html mails. For the activities I think it makes sense to use a ul/li list. Also I think it should be possible to have line breaks in `addBodyText`. What's the best way to do this? I removed all the `htmlspecialchars`, but inside the ul/li the text is then black and not grey anymore, etc.

Should we have a dedicated list method?


MorrisJobke:
> @LukasReschke @MorrisJobke we are currently htmlspecialchars all the input for Html mails. For the activities I think it makes sense to use a ul/li list. Also I think it should be possible to have line breaks in addBodyText. What's the best way to do this? I removed all the htmlspecialchars, but inside the ul/li the text is then black and not grey anymore, etc.

I guess we need to add a new method `addList` to allow such stuff. Because HTML emails are super tricky we need to set the style on each element. 🙈

nickvergessen:
So have addList, which automatically opens ul/li, and closes the list on addFooter/addBodyText again?!

MorrisJobke:
> So have addList, which automatically opens ul/li, and closes the list on addFooter/addBodyText again?!

Something like that ... but it get's quite fast quite complex.

nickvergessen:
> bildschirmfoto vom 2017-04-24 13-46-00

nickvergessen:
Now requires https://github.com/nextcloud/server/pull/4467

MorrisJobke:
Tested and works:

bildschirmfoto 2017-04-25 um 11 27 44

(ignore the purple - that's because of the theming 😉 )

nickvergessen:
Rebased to have the master changes and fixed the unit tests on top of it

PR #142, opened 2017-04-24
add a hint what this settings are

karlitschek:
@nickvergessen @jancborchardt

### DIFF
--- a/templates/settings/form.php
+++ b/templates/settings/form.php
@@ -26,7 +26,13 @@
 
 <form id="activity_notifications" class="section">
 	<h2 class="icon-activity"><?php p($l->t('Activity')); ?></h2>
-	<p class="settings-hint"><?php p($l->t('Here you can configure the default activity settings for users.')); ?></p>
+	<p class="settings-hint">
+		<?php if ($_['setting'] === 'admin') {
+			p($l->t('Here you can configure the default activity settings for new users.'));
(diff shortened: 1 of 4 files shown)


templates/settings/form.php
```diff
@@ -26,6 +26,7 @@
 
 <form id="activity_notifications" class="section">
 	<h2 class="icon-activity"><?php p($l->t('Activity')); ?></h2>
+	<p class="settings-hint"><?php p($l->t('Here you can configure the default activity settings for users.')); ?></p>
```
nickvergessen: This is also used for the personal page, I will make it show a little different line there tomorro.

templates/settings/form.php
```diff
@@ -26,6 +26,13 @@
 
 <form id="activity_notifications" class="section">
 	<h2 class="icon-activity"><?php p($l->t('Activity')); ?></h2>
+	<p class="settings-hint">
+		<?php if ($setting === 'admin') {
+			p($l->t('Here you can configure the default activity settings for new users.'));
+		} else {
+			p($l->t('Here you can configure for which activities you want to receive emails and which should show up in your feed.'));
```
nickvergessen: @karlitschek want to confirm the string?

karlitschek: looks good 👍

PR #151, opened 2017-05-10
Update doc link to 12

nickvergessen:
Moved it closer to the dependency so I hopefully don't forget it next time.


nickvergessen:
Hmm we have redirects in place and since those org urls are out there anyway they need to continue working.

PR #171, opened 2017-06-20
Use PNG icons for emails and ios client

### LINKED ISSUE #169: Clarification on icon handling for events
There seems to be an issue with SVG images in the [activity emails on gmail](https://github.com/nextcloud/server/issues/5075).
IIRC @nextcloud/ios also had issues with SVG, correct?

If so I would go ahead and add a flag to the activity manager, which indicates whether SVG (android and web) or PNG (ios and email) should be used. Then the providers can check the flag before setting the icon.

Other options are:
1. Always PNG (might appear blurry here and there)
2. NC endpoint which returns a PNG while taking a SVG path as a parameter (would require Imagick)
3. Be hated

cc @MorrisJobke @LukasReschke opinions?

nickvergessen:
- [ ] Requires https://github.com/nextcloud/server/pull/5483

Fix #169

### DIFF
--- a/lib/Controller/APIv2.php
+++ b/lib/Controller/APIv2.php
@@ -29,6 +29,7 @@ use OCA\Activity\Exception\InvalidFilterException;
 use OCA\Activity\GroupHelper;
 use OCA\Activity\UserSettings;
 use OCA\Activity\ViewInfoCache;
+use OCP\Activity\IManager;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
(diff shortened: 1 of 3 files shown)


marinofaggiana:
@MorrisJobke is possible for you applicate this feature and the last NC build on https://ios-demo.weasel.rocks ?

MorrisJobke:
> @MorrisJobke is possible for you applicate this feature and the last NC build on https://ios-demo.weasel.rocks ?

Tomorrow earliest - need to wait for the dailies.

PR #182, opened 2017-08-03
Fix image path

nickvergessen:
Caused by nextcloud/server#5682

But the change is sub-optimal so lets see


MorrisJobke:
Server PR is reverted.

PR #184, opened 2017-08-08
Also include type into the index for filters

### LINKED ISSUE #183: Missing index for filters?
E.g. Opening "Security" takes 5-15secs on c.nc.c

by @LukasReschke

nickvergessen:
Fix #183 
### Before
```
Using where
```

### After
```
Using index condition; Using filesort
```


lib/Migration/Version2006000Date20170808155040.php
```diff
@@ -0,0 +1,46 @@
...
+	 * @throws \Doctrine\DBAL\Schema\SchemaException
+	 * @since 13.0.0
+	 */
+	public function changeSchema(IOutput $output, \Closure $schemaClosure, array $options) {
+		/** @var Schema $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('activity');
+		$table->dropIndex('activity_filter_app');
+		$table->addIndex(['affecteduser', 'type', 'app', 'timestamp'], 'activity_filter');
```
LukasReschke: Looks nice what is the impact on updates though. How long does this need on c.nc.c? :)

nickvergessen: 🙊

nickvergessen: @LukasReschke on the copy you gave me (250MB, 500k entries) it took about 7secs.
I guess its okay?

LukasReschke: Yeah. Think so. Then web updater should still be fine and really big instances anyways use CLI :)

PR #186, opened 2017-08-10
Add new option to send mails every 10 minutes

t2d:
I implemented the suggestion from https://help.nextcloud.com/t/email-activity-notification/3543/5


nickvergessen:
As per our documentation the cron system is only run every 15 minutes, so that would be the documented maximum.

However I would recomment to just use: https://www.schilljs.com/2017/05/09/better-scheduling-of-activity-emails.html instead.
With that you can send the emails every x minutes by just running the cron script as prefered.
Not sure a user settings makes sense for this.

Maybe instead of 10 minutes use `As soon as possible` and use a delay of 0?

t2d:
Thank you for the suggestions. I'm going to look into it.

PR #187, opened 2017-08-10
Add option to send mails ASAP aka with next cron run

t2d:
Picking up on https://github.com/nextcloud/activity/pull/186

This PR adds an option to send mails out `As soon as possible`.
For example, you can add a cron job
```
 */5  *  *  *  *    php -f /var/www/nextcloud/occ activity:send-mails asap
```
that will send out the mails of all users which selected `As soon as possible` every 5 minutes.

### DIFF
--- a/lib/MailQueueHandler.php
+++ b/lib/MailQueueHandler.php
@@ -231,6 +231,8 @@ class MailQueueHandler {
 				$query->where($query->expr()->eq('amq_timestamp', $query->createFunction($query->getColumnName('amq_latest_send') . ' + ' . 3600 * 24)));
 			} else if ($restrictEmails === UserSettings::EMAIL_SEND_WEEKLY) {
 				$query->where($query->expr()->eq('amq_timestamp', $query->createFunction($query->getColumnName('amq_latest_send') . ' + ' . 3600 * 24 * 7)));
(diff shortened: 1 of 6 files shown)


lib/MailQueueHandler.php
```diff
@@ -231,6 +231,8 @@ protected function getAffectedUsers($limit, $latestSend, $forceSending, $restric
 				$query->where($query->expr()->eq('amq_timestamp', $query->createFunction($query->getColumnName('amq_latest_send') . ' + ' . 3600 * 24)));
 			} else if ($restrictEmails === UserSettings::EMAIL_SEND_WEEKLY) {
 				$query->where($query->expr()->eq('amq_timestamp', $query->createFunction($query->getColumnName('amq_latest_send') . ' + ' . 3600 * 24 * 7)));
+			} else if ($restrictEmails === UserSettings::EMAIL_SEND_ASAP) {
+				$query->where($query->expr()->eq('amq_timestamp', $query->createFunction($query->getColumnName('amq_latest_send'))));
```
nickvergessen:
```php
$query->where($query->expr()->eq('amq_timestamp', 'amq_latest_send'));
```

is enough

t2d: thank you for pointing this out. I changed the line.

PR #203, opened 2017-09-15
Add an option to disable emails completly

### LINKED ISSUE #122: [feature request] give admin ability to turn off e-mail notifications
If admin doesn't want to set e-mails to users and there is no e-mail settings at server it will be super to give ability to turn off e-mail notifications, so the users couldn't set it.

nickvergessen:
Fix #122

### DIFF
--- a/templates/settings/personal.php
+++ b/templates/settings/personal.php
@@ -20,4 +20,21 @@
  */
 script('activity', 'settings');
 style('activity', 'settings');
-print_unescaped($this->inc('settings/form'));
+/** @var array $_ */
+/** @var \OCP\IL10N $l */
+?>
+
+<form id="activity_notifications" class="section">
+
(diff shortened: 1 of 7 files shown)


templates/settings/personal.php
```diff
@@ -20,4 +20,21 @@
...
+?>
+
+<form id="activity_notifications" class="section">
+
+	<h2><?php p($l->t('Activity')); ?></h2>
+	<p class="settings-hint">
+		<?php if ($_['email_enabled']) { ?>
+			<?php p($l->t('Choose for which activities you want to get an email notification, and which should show up in the activity app stream.')); ?>
+		<?php } else { ?>
+			<?php p($l->t('Choose for which activities you want to see in the your stream.')); ?>
```
MorrisJobke: Remove the `for` 😉

PR #213, opened 2017-10-12
Accessibility fixes

jancborchardt:
Just like the server pull request for accessibility fixes, also fixing some things here. :)

Please review @nextcloud/accessibility @nickvergessen

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -138,6 +138,16 @@
 	font-weight: bold;
 }
 
+.activity-section a:focus,
+.activity-previews a:hover,
+.activity-previews a:focus {
+	opacity: .5;
+}
+
+.activity-section a:hover {
+	opacity: .7;
+}
+
 .activitysubject strong.systemtag {
 	display: inline-block;
 	padding: 2px 5px;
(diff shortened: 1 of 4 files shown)


css/style.css
```diff
@@ -138,6 +138,13 @@
 	font-weight: bold;
 }
 
+.activity-section a:hover,
+.activity-section a:focus,
```
nickvergessen: text looks really awkward now since the grey is very slightly off. Could we use like `.7` which actually matches the normal font color? For the previews the `.5` looks okay

jancborchardt:
> I doubt a little the semantic usefulness of the empty alts, but first, any following analysis runs will not flag them and they definitely don't hurt screen reader accessibility.

@tyrylu according to the guidelines every image should have an alt attribute, and decorative ones should have an empty one. So it’s just confirming they are decorative. :)

css/style.css
jancborchardt: @nickvergessen what do you mean by »slightly off« and not matching the normal font color? .7 is too little difference between the normal state, and keep in mind this is only on hover and focus. :)

nickvergessen: > bildschirmfoto vom 2017-10-13 12-07-14

Yeah the difference between a hovered link and the rest of the text looks awkward I think

jancborchardt: I adjusted the hover opacity to .7 while keeping the rest at .5 :)

Issue #163, opened 2017-06-03
Ongoing PHP errors in /lib/FilesHooks.php during files upload


MichaIng:
Reported by several users/client-configurations on [help.nextcloud.com](https://help.nextcloud.com/t/nc12-beta2-activity-error-php-undefined-offset-3-at-path-to-nextcloud-apps-activity-lib-fileshooks-php-620/12461?u=michaing).

### Steps to reproduce
1. Update to Nextcloud 12, any version since Beta 2
2. Upload files using desktop (>=2.3.0rc1) or android client (1.4.3 at least)

### Expected behaviour
No error messages should be shown in nextcloud log.

### Actual behaviour
Several identical error messages each minute show up in nextcloud log during upload process (single large AND many smaller files) while no actual misbehaviour was observed so far:
`Error	PHP	Undefined offset: 3 at /path/to/nextcloud/apps/activity/lib/FilesHooks.php#620`

### Server configuration
**Operating system**: Raspbian Jessie

**Web server:** Apache/2.4.25

**Database:** MariaDB 10.0.30

**PHP version:** mod_php 7.0.18-1

**Nextcloud version:** Nextcloud 12 Beta 2 until current 12.0.0.29 release

**Where did you install Nextcloud from:** download.nextcloud.com + web updater

**Signing status:** everything passed integrity check

**List of activated apps:**
```
  - activity: 2.5.2
  - calendar: 1.5.3
  - contacts: 1.5.3
  - dav: 1.3.0
  - federatedfilesharing: 1.2.0
  - files: 1.7.2
  - files_downloadactivity: 1.1.1
  - files_sharing: 1.4.0
  - files_trashbin: 1.2.0
  - files_versions: 1.5.0
  - gallery: 17.0.0
  - logreader: 2.0.0
  - lookup_server_connector: 1.0.0
  - nextcloud_announcements: 1.1
  - notifications: 2.0.0
  - oauth2: 1.0.5
  - ownbackup: 17.5.0
  - previewgenerator: 1.0.6
  - provisioning_api: 1.2.0
  - serverinfo: 1.2.0
  - tasks: 0.9.5
  - twofactor_backupcodes: 1.1.1
  - updatenotification: 1.2.0
  - workflowengine: 1.2.0
```

**Nextcloud configuration:**
```
{
    "system": {
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
[...]

PR #226, opened 2017-12-29
Properly construct path of root on file rename

rullzer:
If you have a file in your root folder. And you rename that file. We
will try to find who has access to the parent. In this case the parent
of is the root folder of the user. So sections is only 3 elements big.

Found on our sentry instance.

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -619,7 +619,8 @@ class FilesHooks {
 		$sections = explode('/', $path, 4);
 
 		$accessList['ownerPath'] = '/';
-		if (count($sections) === 4) {
+		if (isset($sections[3])) {
+			// Not the case when a file in root is renamed
 			$accessList['ownerPath'] .= $sections[3];
 		}
 
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -617,7 +617,11 @@ class FilesHooks {
 
 		$path = $node->getPath();
 		$sections = explode('/', $path, 4);
-		$accessList['ownerPath'] = '/' . $sections[3];
+
+		$accessList['ownerPath'] = '/';
+		if (count($sections) === 4) {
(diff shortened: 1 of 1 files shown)


lib/FilesHooks.php
```diff
@@ -617,7 +617,11 @@ protected function getUserPathsFromPath($path, $uidOwner) {
 
 		$path = $node->getPath();
 		$sections = explode('/', $path, 4);
-		$accessList['ownerPath'] = '/' . $sections[3];
+
+		$accessList['ownerPath'] = '/';
+		if (count($sections) === 4) {
```
MorrisJobke: I would use an `isset($sections[3])` because it makes semantically more sense (4+ elements is fine and it actually tests, the element that is actually used later - `[0 => 'a', 1 => 'b', 2 => 'c', 4 => 'd']` is also an array of the length 4 😉  (even if it is not typical to be the case here)

rullzer: https://github.com/nextcloud/activity/pull/226/files/fbd374d889459e7148b6965cfb0d9d6690738e6a#diff-dbc5f6ba450710b214d9d73b15c9f186R619 it is max 4 elements...

rullzer: but sure

MorrisJobke: I know, this is just about better practises and easier to understand code snippets.

nickvergessen: Sounds good, but yeah isset on 3 seems more logical.

Thanks for finding the issue.

nickvergessen:
Should backport to 12?
https://github.com/nextcloud/activity/blob/stable12/lib/FilesHooks.php#L619-L620

nickvergessen:
Or is it just a php notice?

MorrisJobke:
@nickvergessen Looking at #163 I would backport it to 12.

PR #230, opened 2018-01-10
Add viewbox to svgs to avoid sizing issues in IE

juliushaertl:
Before:
bildschirmfoto vom 2018-01-10 09-40-48 kopie

After:
bildschirmfoto vom 2018-01-10 09-44-18 kopie

PR #231, opened 2018-01-10
[stable12] Add viewbox to svgs to avoid sizing issues in IE

juliushaertl:
backport of #230


MorrisJobke:
Drone only does the sign-off checks: merging

PR #247, opened 2018-02-15
ensure userids are always strings

blizzz:
Fixes an ``"InvalidArgumentException\",\"Message\":\"The given affected user is invalid\"``issue.

Repro steps:

1. Have two users with numerical ideas, "12345" and "67890"
2. Let those two users have a shared folder, containing a subfolder and a file
3. Move  the file into the subfolder

Without the patch, the exception ↑ is logged.

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -1056,6 +1056,7 @@ class FilesHooks {
 			return;
 		}
 
+		$user = (string)$user;
 		$selfAction = $user === $this->currentUser->getUID();
 		$app = $type === Files_Sharing::TYPE_SHARED ? 'files_sharing' : 'files';
 		$link = $this->urlGenerator->linkToRouteAbsolute('files.view.index', array(


nickvergessen:
Since they all call `addNotificationsForUser`in the end, casting there should be enough?!

blizzz:
> Since they all call addNotificationsForUserin the end, casting there should be enough?!

Yes, would work, too. My approach was to address this earlier and have the ids where possible in the proper type, so that they are passed around correctly, even if stuff changes later further down the paths.

blizzz:
All in all I am also not that fond and would prefer to ship IUser around instead, as this is fragile and especially app devs can fall into this trap easily. That would not be backportable, though. In that sense, casting it at the `setAffectedUser()` call would suffice, if it was guaranteed to fix this for 14.

blizzz:
@nickvergessen changed to the single cast

blizzz:
backports already in https://github.com/nextcloud/activity/pull/252 and https://github.com/nextcloud/activity/pull/253

PR #263, opened 2018-04-16
Vue.js

nickvergessen:
- [x] Use Vue.js
- [x] Use `vue-routing`
- [ ] Use `vue-localstorage`
- [ ] Reload data from local storage when switching back to a filter that was visited before
- [ ] Infinite loading should use `vue-infinite-loading`


MorrisJobke:
@nickvergessen Moved to 15.

MorrisJobke:
@nickvergessen What is the status here? We are close to the beta 1. I guess this will go into 18, right?

skjnldsv:
I would say that localstorage is secondary.
With a proper loading it should already be faster in vue.
We could get this pr in and then focus on the local store? :)

nickvergessen:
I'm pretty sure in my memory this PR does not fully work. But since I'm drawning in Talk tasks, anyone can take over this simple app :P

Issue #269, opened 2018-05-10
Allow Activity registers actions of share and unshare for circles


fgsl:
### Steps to reproduce
1. Install Nextcloud 
2. Install Activity App
3. Install Circles App

### Expected behaviour
When a file is shared or unshared for a circle, Activity view must show these actions.

### Actual behaviour
Activity view doesn't show actions of share and unshare for circles.

**Nextcloud version:** 14
**Activity version:** 2.6.1
**Circles version:** 0.13.6

**Updated from an older Nextcloud/ownCloud or fresh install:** Fresh install.

**Where did you install Nextcloud from:** Github master.

It's clear that actions won't be registered if Circles app doesn't emit signals for sharing and unsharing, but this issue will be sent to backlog of Circles. Needed changes in files_sharing app already was sent by https://github.com/nextcloud/server/issues/9451.

nickvergessen:
> That should be done by the circles app ;)
Please report to https://github.com/nextcloud/circles/

fgsl:
Hi, @nickvergessen .

I studied how to do this in circles app and think that is not the right place.

If I create a `FilesHooks` class in circles app to add notifications for user, there will be a moment when it must be necessary to use activity implementation and so to create a high coupling between apps.

This change in activity `FilesHooks` class does not create coupling with circles, because there are no circles namespaces.

If I do what you suggests, I will use activity namespaces in circles. My change does not add new namespaces, in fact I use server namespace (`Share` class).

nickvergessen:
No, you can get the activity manager via `\OC::$server->getActivityManager()` in the circles app and then do everything you need to do ;)

PR #270, opened 2018-05-10
Allow Activity registers actions of share and unshare for circles #269

fgsl:


nickvergessen:
That should be done by the circles app ;)
Please report to https://github.com/nextcloud/circles/

PR #302, opened 2018-10-16
remove unneeded -ms-filter (was only needed for IE8)

coliff:


css/style.css
```diff
@@ -195,7 +195,6 @@
 	margin-top: 2px;
 	background-size: contain;
 
-	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=50)";
 	filter: alpha(opacity=50);
```
juliushaertl: We also don't need the filter: alpha... anymore, can you remove that as well?

juliushaertl:
Looks good now :+1: @coliff Can you squash your commits into one and add a sign-off message?

coliff:
@juliushaertl  Sorry - I’m not near a computer for a little while and I don’t think I can do that from the web interface. Can you do that?

PR #311, opened 2018-11-11
Correctly restrict affected users when using command to send emails

mlojewski-me:
When invoking 'occ activity:send-mails <restrict-batching>' correctly download affected users: amq_timestamp = amq_latest_send - batchtime

### DIFF
--- a/lib/MailQueueHandler.php
+++ b/lib/MailQueueHandler.php
@@ -231,11 +231,11 @@ class MailQueueHandler {
 
 		if ($restrictEmails !== null) {
 			if ($restrictEmails === UserSettings::EMAIL_SEND_HOURLY) {
-				$query->where($query->expr()->lte('amq_timestamp', $query->createFunction($query->getColumnName('amq_latest_send') . ' + ' . 3600)));
+				$query->where($query->expr()->eq('amq_timestamp', $query->createFunction($query->getColumnName('amq_latest_send') . ' - ' . 3600)));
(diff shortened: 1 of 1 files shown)


lib/MailQueueHandler.php
```diff
@@ -231,11 +231,11 @@ protected function getAffectedUsers($limit, $latestSend, $forceSending, $restric
 
 		if ($restrictEmails !== null) {
 			if ($restrictEmails === UserSettings::EMAIL_SEND_HOURLY) {
-				$query->where($query->expr()->lte('amq_timestamp', $query->createFunction($query->getColumnName('amq_latest_send') . ' + ' . 3600)));
+				$query->where($query->expr()->eq('amq_timestamp', $query->createFunction($query->getColumnName('amq_latest_send') . ' - ' . 3600)));
```
nickvergessen: ah, the lte made it execute all emails.

MorrisJobke: I still don't get it. What if the batch run is one second late. Then it does not execute anymore for those users. 🤔

mlojewski-me: The time of batch run does not matter. `amq_timestamp` is time of an activity event, `amq_latest_send` is always `amq_timestamp` plus seconds that user set for its mail notification interval, so query gets only users who have any notifications and these notifications' `amq_timestamp = amq_latest_send - batchtime`.

MorrisJobke: Got it 👍 I misread the `amq_timestamp` and `amq_latest_send` as the same name 🙈

PR #316, opened 2018-11-12
Use the full path instead of the file name in HTML emails too

nickvergessen:
Fix https://github.com/nextcloud/server/issues/12309


MorrisJobke:
@nickvergessen Pending backport?

PR #327, opened 2018-12-10
Added Webhooks methods to complement Stream and E-Mail notifications.

tinko92:
The commit in this pull request adds the feature to send HTTP(S) requests to a URL that is specified in the admin settings as a notification for new activities, similar to the existing methods of sending notifications about new activities via stream or e-mail. The request contains information about the event in JSON format and optionally a secret string that can be set in the Admin settings and can be verified on the receiving end.

Signed-off-by: Tinko Bartels <mail@tinkobartels.de>

### DIFF
--- a/css/settings.css
+++ b/css/settings.css
@@ -25,3 +25,11 @@ table.activitysettings td.small label {
 .nav-icon-activity {
 	background-image: url('../img/activity-dark.svg?v=1');
 }
+
+#webhook_settings_form {
+	display:inline-block;
+}
+
+#webhook_url {
+	width: 260px;
+}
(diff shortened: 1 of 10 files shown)


MatthiasHertel:
@nickvergessen @icewind1991 

hey why not merged ?
i think it would be nice to register more notification provider (slack webhook) for give a possiblity preventing load on mail server.

can u review it and merge ? 

it would be nice :)

nickvergessen:
It might be revisited soon, as there are plans to change activity and notifications in something new

nickvergessen:
Closing this as in it's current state this is too far from being mergable.

* Curl is used instead of Nextcloud's HTTP client, allowing SSRF
* Webhook results are not signed with a shared secrect, so can not be validated on the recipient side for authenticy, instead the token is send directly as header
* Settings were redone

A follow up approach should be similar to https://github.com/nextcloud/spreed/pull/9458

PR #346, opened 2019-02-06
Don't add empty entries to the objects array

nickvergessen:


tobiasKaminsky:
Will this be backported? It is quite a problem on Android if no activities at all are shown.

nickvergessen:
Backport doesnt help as releases just went out and it would be broken 4-6 more weeks.

tobiasKaminsky:
Then I'll add a fix to Android :+1:

PR #351, opened 2019-02-19
Only apply .box and .group class styles if they are related to activity

juliushaertl:
`.group` is also used in the files sharing dialog, so the css rules should only apply if they are activity related.

Found when testing https://github.com/nextcloud/server/pull/14180

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -52,72 +52,74 @@
 	border-radius: 5px;
 }
 
-.group {
+.activity-section .group {
 	clear: both;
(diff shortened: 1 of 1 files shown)


nickvergessen:
Fine by me, I guess the whole file should receive an overhole after #263

nickvergessen:
Backport to 15? Or is it only a problem on master?

juliushaertl:
Doesn't happen on 15 for me, so should be fine on master only.

PR #357, opened 2019-03-28
Fallback to plaintext subject if no rich subject is set

nickvergessen:
Fix https://github.com/nextcloud/server/pull/14844


nhirokinet:
Is it okay that the text which was plain-text will be returned without `htmlspecialchars`?

nickvergessen:
Of course it's not. Fixed

PR #366, opened 2019-04-16
Drop activity_time index

rullzer:
To avoid other queries prefering this index and being painfully slow.

As discussed.


nickvergessen:
Can you also comment out the place were it's added. No need to punish late updater and new installers with it.

rullzer:
Done

rullzer:
Backport to 16?

PR #369, opened 2019-04-18
Remove deprecated and old update file

nickvergessen:


nickvergessen:
The file is not needed anymore since a couple of releases. Since you cant update directly anymore, we can remove it.

PR #379, opened 2019-06-21
Fix "unshare group share from self" activity

nickvergessen:
Fix https://github.com/nextcloud/server/issues/15759


nickvergessen:
lets ignore 14 for now

PR #383, opened 2019-08-12
Add expiration event for shares

skjnldsv:
Needs https://github.com/nextcloud/server/pull/16723

| 1 | 2 |
|:---------:|:------:|
|Capture d’écran_2019-08-12_11-07-35|Capture d’écran_2019-08-12_11-07-43|

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -793,17 +793,27 @@ class FilesHooks {
 			return;
 		}
 
+		if ($share->isExpired()) {
+			$actionSharer = 'expired_user';
+			$actionOwner = 'expired_user';
+			$actionUser = 'expired';
+		} else {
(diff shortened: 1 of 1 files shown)


lib/FilesHooks.php
```diff
@@ -793,17 +793,28 @@ protected function unshareFromUser(IShare $share) {
 			return;
 		}
 
+		if ($share->isExpired()) {
+			// Link expired
+			$actionSharer = 'expired_user_by';
```
nickvergessen: actually I think the `_by` is confusing and should be removed. It's not expired by the user, but by time...

nickvergessen: This is true for all expire messages

skjnldsv: Good to me!

PR #395, opened 2019-09-21
Don't overwrite user settings on login

PhrozenByte:
Using `notify_setting_batchtime` to detect whether all app settings of a particular user have been set already isn't the best solution, since it makes it impossible to change a user's app settings (e.g. via CLI) before his first login. This commit introduces a new (hidden) `configured` setting to check whether all settings have been set already. If they haven't, they are set, but existing settings won't be overwritten.

### DIFF
--- a/lib/Hooks.php
+++ b/lib/Hooks.php
@@ -72,7 +72,7 @@ class Hooks {
 
 	static public function setDefaultsForUser($params) {
 		$config = \OC::$server->getConfig();
-		if ($config->getUserValue($params['uid'], 'activity', 'configured', null) !== null) {
+		if ($config->getUserValue($params['uid'], 'activity', 'configured', 'no') === 'yes') {
 			// Already has settings
 			return;
 		}
@@ -95,8 +95,8 @@ class Hooks {
 			);
 		}
 
-		// Mark settings as set
-		$config->setUserValue($params['uid'], 'activity', 'configured', '1');
+		// Mark settings as configured
+		$config->setUserValue($params['uid'], 'activity', 'configured', 'yes');
 	}
 
 	/**
--- a/lib/Hooks.php
+++ b/lib/Hooks.php
@@ -72,13 +72,18 @@ class Hooks {
 
 	static public function setDefaultsForUser($params) {
 		$config = \OC::$server->getConfig();
-		if ($config->getUserValue($params['uid'], 'activity','notify_setting_batchtime', null) !== null) {
+		if ($config->getUserValue($params['uid'], 'activity', 'configured', null) !== null) {
 			// Already has settings
 			return;
 		}
 
 		foreach ($config->getAppKeys('activity') as $key) {
 			if (strpos($key, 'notify_') !== 0) {
+				continue;
+			}
+
+			if ($config->getUserValue($params['uid'], 'activity', $key, null) !== null) {
+				// Already has this setting
 				continue;
 			}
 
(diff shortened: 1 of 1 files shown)


nickvergessen:
So why do you set the stuff manually on the database, instead of changing the default settings in the admin interface?

PhrozenByte:
We're importing users via CLI and depending on the user's group the default settings are different.

PhrozenByte:
Any updates yet @nickvergessen?

nickvergessen:
And what about if you also set `'notify_setting_batchtime'` that should work just as fine?

PhrozenByte:
1. I'd have to set all options from the beginning, not just `notify_setting_batchtime` and the one I want to have a different default value. It additionally forces me to strictly monitor this app's development, as it may add or modify config parameters at any time. It's also repetitive: I want to change the defaults of a single setting (behaviour of mail notifications to be more precise) only, I don't want to manage the default settings at multiple places, making space for mistakes.

2. Speaking of new config parameters: Something like a `configured` key is a prerequisite for adding new config parameters, at least unless you still want to leave the decision about its default value for existing users up to the admin. In other words: It allows a smooth transition of configs.

3. One simply doesn't expect `notify_setting_batchtime` to act like an `configured` key. Since there's an official API to set config params via CLI, I should be able to expect the config params to to what they tell me to do - i.e. that `notify_setting_batchtime` does exactly one thing. Currently it's acting like a magic number. I must read the code to know this. I actually stumbled over this, this undocumented behaviour was the reason why I checked the code.

4. It's a very simple change that has, at least as far as I can assess, zero negative side effects. Could you please elaborate your concerns? They don't seem obvious to me, but I definitely don't know enough about this app.

nickvergessen:
Well teh advantage would be no modification is needed.

But you are right, it's a bit of magic involved.

lib/Hooks.php
```diff
@@ -72,7 +72,7 @@ static protected function deleteUserMailQueue(IDBConnection $connection, $user)
 
 	static public function setDefaultsForUser($params) {
 		$config = \OC::$server->getConfig();
-		if ($config->getUserValue($params['uid'], 'activity','notify_setting_batchtime', null) !== null) {
+		if ($config->getUserValue($params['uid'], 'activity', 'configured', null) !== null) {
```
nickvergessen:
```suggestion
		if ($config->getUserValue($params['uid'], 'activity', 'configured', 'no') === 'yes') {
```

lib/Hooks.php
```diff
@@ -82,13 +82,21 @@ static public function setDefaultsForUser($params) {
...
 			$config->setUserValue(
 				$params['uid'],
 				'activity',
 				$key,
 				$config->getAppValue('activity', $key)
 			);
 		}
+
+		// Mark settings as set
+		$config->setUserValue($params['uid'], 'activity', 'configured', '1');
```
nickvergessen:
```suggestion
		$config->setUserValue($params['uid'], 'activity', 'configured', 'yes');
```

lib/Hooks.php
```diff
@@ -82,13 +82,21 @@ static public function setDefaultsForUser($params) {
...
+
 			$config->setUserValue(
 				$params['uid'],
 				'activity',
 				$key,
 				$config->getAppValue('activity', $key)
 			);
 		}
+
+		// Mark settings as set
```
nickvergessen:
```suggestion
		// Mark settings as configured
```

nickvergessen:
Lets remove some more magic and then this can go in

PhrozenByte:
Even though there's this nice "commit patch" feature in the GitHub web UI, GitHub still can't sign off commits, requiring a force push :laughing: 

All fixed, thank you @nickvergessen :+1:

nickvergessen:
Thanks fopr the contribution, I will take core of the failing ci tomorrow

PR #406, opened 2019-11-20
is_dir can be null on blacklisted files

kokspflanze:
if you set blacklisted_files in your config like

````
'blacklisted_files' => [
    '.htaccess',
    'Thumbs.db',
    'thumbs.db',
],
````

than is the `is_dir` parameter null but the `getPreviewLink` method require a bool, so it generate a error and the activity page hangs.

### DIFF
--- a/lib/Controller/APIv2.php
+++ b/lib/Controller/APIv2.php
@@ -357,7 +357,7 @@ class APIv2 extends OCSController {
 		}
 
 		$preview = [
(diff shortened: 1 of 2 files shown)


nickvergessen:
Can you instead add a casting to boolean in the `ViewInfoCache` class when we set the value?
Thanks a lot

kokspflanze:
@nickvergessen i was thinking about this too, but i was thinking null is needed to know if a file is blacklisted.

but i will change it

PR #421, opened 2020-01-25
Fix for NC Issue 18695

Dennis1993:
We can fix this issue/idea by bring in a simple line. If we move the timestamp a litte bit more to the top then we have a problem with long entries which make the timestamp not readable.

Thats my idea

image

https://github.com/nextcloud/server/issues/18695

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -53,6 +53,8 @@
 .activity.box {
 	width: 100%;
 	display: block;
+	border-bottom: 1px solid var(--color-border-dark);
+	padding-bottom: 10px;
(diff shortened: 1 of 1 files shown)


nickvergessen:
I think instead we should adjust the position of the timestamp

artonge:
Not necessary anymore as of the vue rewrite from a few month ago.

artonge:
hello@josandcamila.com

artonge:
Not necessary anymore as of the vue rewrite of a few month ago.

PR #433, opened 2020-03-02
Chunk deleting of rows from the activity table

### LINKED ISSUE #432: Chunk deleting of rows from the activity table
We are using galera cluster and hitting wsrep_max_ws_rows during deleteActivities because of large raw count.

Is it possible to take over the fix from owncloud for this issue into the nextcloud code base?
https://github.com/owncloud/activity/commit/84e6f269019583fa123e4d89e0ac3104f5b75a5f

DonCarlosGraz:
fixes #432 
see also https://github.com/owncloud/activity/commit/84e6f269019583fa123e4d89e0ac3104f5b75a5f

### DIFF
--- a/lib/Data.php
+++ b/lib/Data.php
@@ -24,6 +24,7 @@
 
 namespace OCA\Activity;
 
+use Doctrine\DBAL\Platforms\MySqlPlatform;
 use OCP\Activity\IEvent;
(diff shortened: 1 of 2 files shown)


nickvergessen:
Limit on delete is not supported apart from mysql

PR #436, opened 2020-03-16
Harden code to not explode if a storage has no owner

rullzer:
Related to https://github.com/nextcloud/server/pull/19978

Signed-off-by: Roeland Jago Douma <roeland@famdouma.nl>

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -635,7 +635,11 @@ class FilesHooks {
 	 */
 	protected function getSourcePathAndOwner($path) {
 		$view = Filesystem::getView();
-		$owner = $view->getOwner($path);
+		try {
+			$owner = $view->getOwner($path);
+		} catch (NotFoundException $e) {
+			$owner = null;
+		}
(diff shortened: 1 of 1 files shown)


nickvergessen:
I'd like to not do this and instead fix storages to always return an owner. and if they dont have one, the mountpoint owner should be returned in my opinion:
https://github.com/nextcloud/groupfolders/pull/797

nickvergessen:
> Conflicting files
lib/FilesHooks.php

:speak_no_evil: 

> I would still do this as then we would not break existing storage implementations that do not provide an owner yet.

Well, yes and no. Not breaking is good. On the other hand this is the only point where we know that our storages still return wrong info

rullzer:
so... merge or close?

juliushaertl:
> Well, yes and no. Not breaking is good. On the other hand this is the only point where we know that our storages still return wrong info.

Fine by me then, we still have some time to see during 19 beta/rc if that breaks somewhere and fix it then.

PR #448, opened 2020-04-04
Skip notifications for users with invalid email address.

kesselb:
Fix https://github.com/nextcloud/server/issues/20294 https://github.com/nextcloud/server/issues/15600

How to test:

- Have Alice (email: alice@nextcloud.test) and Bob (email: bob).
- Both are members of group Test
- Share a folder to group Test
- Run `occ activity:send-mails`
- We are able to send the notification for Alice but for Bob it fails with something like

image

- Run `occ activity:send-mails` again
- See the notification for Alice is sent a second time

The faulty code is actual at https://github.com/nextcloud/server/blob/1a9330cd69631ef12b71239c4f98836e09e021ea/lib/private/Mail/Message.php#L82-L83. We might fix that by just deleting `Message::convertAddresses` and `Mailer::convertEmail` method because SwiftMailer supports IDN nowdays.

### DIFF
--- a/lib/MailQueueHandler.php
+++ b/lib/MailQueueHandler.php
@@ -325,6 +325,11 @@ class MailQueueHandler {
 			return true;
 		}
 
+		if (!$this->mailer->validateMailAddress($email)) {
+			$this->logger->warning('Notification for user "{user}" not sent because the email address "{email}" is invalid.', ['user' => $userName, 'email' => $email]);
+			return true;
+		}
+
 		list($mailData, $skippedCount) = $this->getItemsForUser($userName, $maxTime);
 
 		$l = $this->getLanguage($lang);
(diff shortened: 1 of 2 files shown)


lib/MailQueueHandler.php
```diff
@@ -325,6 +325,11 @@ protected function sendEmailToUser($userName, $email, $lang, $timezone, $maxTime
 			return true;
 		}
 
+		if (!$this->mailer->validateMailAddress($email)) {
+			$this->logger->warning('Notification for user "{user}" not sent because the email address "{email}" is invalid.', ['user' => $userName, 'email' => $email]);
+			return true;
```
nickvergessen, line 330: should we really delete them? or maybe just wait and send them once it's fixed?

kesselb: Don't know :) Usually the addresses are validated. Mind to ask the ldap / saml experts? It's more likely to happen with such setups.

kesselb: I choosed return true because a invalid email address is nothing temporary. Without manual action it will still fail with the next run.

nickvergessen: Well it could then send the emails once the admin fixed it. But welli don't care too much

PR #460, opened 2020-05-06
add debug output and missing exception handlers to fix #459

### LINKED ISSUE #459: Notification are sent over and over again
I reported this already to https://github.com/nextcloud/server/issues/20286 before I realized now that it seems to be cause by this app. Please refer to that issue for details.

Trying to find the cause, I could find out that the events are not deleted from the events queue in "activity_mq" table.

Also, "amq_last_send" is not updated, it's always the same as "amq_timestamp".

Finally I cloud solve this after adding some debug statements to MailQueueHandler.php . I will extend this shortly.

madmas:
add debug outout dan exception handling fir to get rid of over-and-over sent notifications.


kesselb:
How does logException fixes something? Mind to share `$e->getMessage()` to see why sending the email failed?

madmas:
Hi @kesselb , sorry, I switched the lines, I think the actual fix is below, in line 415. 410 of course only helps in regard of debugging ;-)

kesselb:
Mind to share the exception? Also you overwrite the actual exception message. That makes it harder to debug.

lib/MailQueueHandler.php
```diff
@@ -132,6 +132,7 @@ public function sendEmails($limit, $sendTime, $forceSending = false, $restrictEm
 			// No users found to notify, mission abort
 			return 0;
 		}
+		$this->logger->debug('Sending notification emails to users: '. implode("|",$affectedUsers));
```
kesselb, line 135:
```suggestion
		$this->logger->debug('Sending notification emails to users: '. implode(',', $affectedUsers));
```

lib/MailQueueHandler.php
```diff
@@ -177,6 +178,7 @@ public function sendEmails($limit, $sendTime, $forceSending = false, $restrictEm
 		}
 		$this->activityManager->setRequirePNG(false);
 
+		$this->logger->debug('eMail notifications sent. Will delete: ' . implode("|",$deleteItemsForUsers));
```
kesselb, line 181:
```suggestion
		$this->logger->debug('Email notifications sent. Will delete: ' . implode(', ', $deleteItemsForUsers));
```

lib/MailQueueHandler.php
```diff
@@ -405,6 +407,12 @@ protected function sendEmailToUser($userName, $email, $lang, $timezone, $maxTime
 		try {
 			$this->mailer->send($message);
 		} catch (\Exception $e) {
+			$this->logger->logException($e, [
+							'message' => 'Failed sending email to user "{user}"',
+							'user' => $user,
```
kesselb, line 412: Log the exception with `$this->logger->logException($e);` and log a additional warning `$this->logger->warning('Failed to send notification email for user {user}', ['user' => $userName]);`. You cannot use `$user` because it's a IUser object.

madmas:
> I'm not convinced. ~Right now the error handling code fails and~ the original exception message is replaced with a generic error message.

My experience was vice versa - with the output I added with shown here, I got much more details.
I cannot share the exception as after implementing this, I don't see them happen on my system anymore....

kesselb:
We still don't know why sending the notification failed in the first place. This is fixing symptoms not problems 😟

PR #467, opened 2020-06-16
Move to Vue

ChristophWurst:
Todo:
- [x] fetch and display filters
- [ ] Show icon for filters
- [ ] Show the default filter by default
- [ ] Show proper filter on selection
- [ ] Parse and display activities
- [ ] Make filtering to generic that we can also pass in a type + id (sidebar)
- [ ] Invinite scrolling for activities


ChristophWurst:
stale and conflicting -> close

PR #481, opened 2020-08-11
merge file activity settings into a single 'favorite changed' item

icewind1991:
see https://github.com/nextcloud/server/pull/22205

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -26,6 +26,7 @@ namespace OCA\Activity;
 
 use OC\Files\Filesystem;
 use OC\Files\View;
+use OC\Tags;
 use OCA\Activity\BackgroundJob\RemoteActivity;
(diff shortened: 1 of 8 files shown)


lib/FilesHooks.php
```diff
@@ -145,45 +153,57 @@ public function fileCreate($path) {
...
 	}
 
 	/**
 	 * Store the delete hook events
+	 *
 	 * @param string $path Path of the file that has been deleted
 	 */
 	public function fileDelete($path) {
-		$this->addNotificationsForFileAction($path, Files::TYPE_SHARE_DELETED, 'deleted_self', 'deleted_by');
+		$this->addNotificationsForFileAction($path, 'deleted_self', 'deleted_by');
```
nickvergessen, line 177: this breaks the non-monochrome activity icon handling across all clients, which is why I had pushed a revert on that in previous PRs.

PR #483, opened 2020-08-14
add daily activity digest option

icewind1991:
- Per user option to enable digest mails
- Digest includes up to 20 activity items from the last day
- If more then 20 activity items exist for the last day, "and %n more" will be added to the mail.
- Email format based on the notification mails:

image


rullzer:
@jancborchardt should we also here filter out your own actions?

nickvergessen:
This can be achieved already via https://docs.nextcloud.com/server/16/admin_manual/configuration_server/activity_configuration.html#better-scheduling-of-activity-emails

But yeah then it affects all users, but we have less duplicate code?

rullzer:
> This can be achieved already via https://docs.nextcloud.com/server/16/admin_manual/configuration_server/activity_configuration.html#better-scheduling-of-activity-emails
> 
> But yeah then it affects all users, but we have less duplicate code?

Always good!

jancborchardt:
> @jancborchardt should we also here filter out your own actions?

Yes – any kind of _notification_ should always filter out anything you do yourself.

(The "Activity" list doesn’t notify, so it’s fine to show your own actions there.)

lib/DigestSender.php
```diff
@@ -0,0 +1,215 @@
...
+		);
+		$skippedCount = max(0, $count - self::ACTIVITY_LIMIT);
+
+		$template = $this->mailer->createEMailTemplate('activity.Notification', [
+			'displayname' => $user->getDisplayName(),
+			'url' => $this->urlGenerator->getAbsoluteURL('/'),
+			'activityEvents' => $activities,
+			'skippedCount' => $skippedCount,
+		]);
+		$template->setSubject($l10n->t('Daily activity digest for ' . $this->defaults->getName()));
```
jancborchardt, line 144: Better wording:
> Daily Nextcloud activity summary

Just to be clear – this is about activity, not only notifications, right? Otherwise it should say "Daily Nextcloud notification summary" of course.

lib/DigestSender.php
```diff
@@ -0,0 +1,215 @@
...
+		$template->addHeader();
+		$template->addHeading($l10n->t('Hello %s', [$user->getDisplayName()]), $l10n->t('Hello %s,', [$user->getDisplayName()]));
```
jancborchardt, line 146: This introduction is a bit useless, let’s instead repeat the subject line ("Nextcloud" not needed as the logo is above):
> Daily activity summary

lib/DigestSender.php
```diff
@@ -0,0 +1,215 @@
...
+
+		$homeLink = '<a href="' . $this->urlGenerator->getAbsoluteURL('/') . '">' . htmlspecialchars($this->defaults->getName()) . '</a>';
+		$template->addBodyText(
+			$l10n->t('There was some activity at %s', [$homeLink]),
+			$l10n->t('There was some activity at %s', [$this->urlGenerator->getAbsoluteURL('/')])
```
jancborchardt, line 151: The body text is quite wishy-washy, we can just cut it.

templates/settings/personal.php
```diff
@@ -37,4 +37,9 @@
 
 	<?php print_unescaped($this->inc('settings/form')); ?>
 
+	<h2><?php p($l->t('Activity')); ?></h2>
+
+	<input id="activity_email_enabled" name="activity_digest" type="checkbox" class="checkbox"
+		   value="1" <?php if ($_['activity_digest_enabled']) { print_unescaped('checked="checked"'); } ?> />
+	<label for="activity_email_enabled"><?php p($l->t('Send daily activity digest')); ?></label>
```
jancborchardt, line 44: → "Send daily activity summary"

lib/DigestSender.php
icewind1991: > this is about activity, not only notifications

yes

icewind1991:
> Is it possible to get avatars and file previews here instead of only the updated/created/deleted icons?

that's would be a fair amount of work to properly handle authentication of the preview urls, I'll look into it in a follow up PR if there is enough time.

lib/DigestSender.php
```diff
@@ -71,21 +71,41 @@ public function __construct(
...
+				if ($timezoneDate->format('H') < 6) {
+					// Still before 6am, so dont send yet.
+					$timezoneDate->sub(new \DateInterval('P1D'));
+				}
+				$timezoneDigestDay[$timezone] = $timezoneDate->format('Y.m.d');
+			}
+
+			$userDigestDate = $digestDate[$user] ?? '';
+			if ($userDigestDate === $timezoneDigestDay[$timezone]) {
+				// User got todays digits already
```
MorrisJobke, line 103:
```suggestion
				// User got todays digest already
```

nickvergessen (approved):

PR #490, opened 2020-09-05
fix new event documentation

sualko:


docs/create.md
```diff
@@ -8,22 +8,22 @@ To create and publish an event to the activity app, a new `IEvent` should be fet
...
 * `setAffectedUser()`
-* `setAuthor()`
-* `setTimestamp()`
 * `setSubject()`
-* `setObject()`
 
 Additionally these values **can** be set:
+* `setAuthor()` - if no author is set, the current user will be used
+* `setTimestamp()` - if no not set, the current time will be used
+* `setObject()`
```
nickvergessen, line 26: This is just not enforced to break old activities, but it should be in the list above so we can at somepoint say "okay ignore all items without objects, the future is here"

nickvergessen (approved):
Looks good in general, apart from the setObject change

sualko:
This pr just reflects the code documentation: https://github.com/nextcloud/server/blob/f91a85c32eb4a4c3cef20e0daab4069437df5d07/lib/public/Activity/IManager.php#L38-L51 

```php
        /**
	 * Generates a new IEvent object
	 *
	 * Make sure to call at least the following methods before sending it to the
	 * app with via the publish() method:
	 *  - setApp()
	 *  - setType()
	 *  - setAffectedUser()
	 *  - setSubject()
	 *
	 * @return IEvent
	 * @since 8.2.0
	 */
	public function generateEvent(): IEvent;
```

So if `setObject` should be mandatory, at lest the code should be updated as well. Btw I have no clue what the difference is between all those functions. I can only make an reasonable assumption based on other apps, but I'm not sure what is best practice and how should it be used. Therefore some documentation from someone with more knowledge would be helpful.

nickvergessen:
Ah oh... hmm then let me fix server and you move setObject() back here

nickvergessen (approved):

PR #500, opened 2020-09-23
Notify groupmates also when renaming and moving files (by using Group folders app)

chrisfritsche:
Currently:
Groupmates (using Group folders app) are only notified when a file is created.

With this PR:
Notify groupmates also when renaming and moving.

Thanks in advance for a feedback.

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -226,18 +226,7 @@ class FilesHooks {
 
 		$this->generateRemoteActivity($accessList['remotes'], $activityType, time(), $this->currentUser->getCloudId(), $accessList['ownerPath']);
 
-		if ($this->config->getSystemValueBool('activity_use_cached_mountpoints', false)) {
-			$mountsForFile = $this->userMountCache->getMountsForFileId($fileId);
-			$affectedUserIds = array_map(function (ICachedMountInfo $mount) {
-				return $mount->getUser()->getUID();
-			}, $mountsForFile);
-			$affectedPaths = array_map(function (ICachedMountFileInfo $mount) {
-				return $this->getVisiblePath($mount->getPath());
-			}, $mountsForFile);
-			$affectedUsers = array_combine($affectedUserIds, $affectedPaths);
-		} else {
-			$affectedUsers = $accessList['users'];
-		}
+        $affectedUsers = $this->getAffectedUsers($accessList['users'], $fileId);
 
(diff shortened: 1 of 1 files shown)


chrisfritsche:
Hi @nickvergessen !
Pls when you have time, maybe you can also review this PR for me? Thanks a lot in advance.

lib/FilesHooks.php
```diff
@@ -461,8 +450,8 @@ protected function fileMoving($oldPath, $newPath) {
 			return;
 		}
 		$accessList = $this->getUserPathsFromPath($parentPath, $parentOwner);
-		$affectedUsers = $accessList['users'];
-		$oldUsers = $this->oldAccessList['users'];
+		$affectedUsers = $this->getAffectedUsers($accessList['users'], $fileId);
+		$oldUsers = $this->getAffectedUsers($this->oldAccessList['users'], $fileId);
```
nickvergessen, line 454: I think you need to do this where `$this->oldAccessList` is written, because `getMountsForFileId` and so on might not work now anymore. So you can not get people that had access before the action

chrisfritsche: @nickvergessen  Thx for the feedback!
So pls correct me if i understood wrong: In this case, we can just leave this line as before:
Because the new users will be notified (included in the previous line) and the old users, that loose the visibility to the file, don't need to be no notified. Correct?

nickvergessen: Search for:
```php
$this->oldAccessList = $this->getUserPathsFromPath($this->oldParentPath, $this->oldParentOwner);
```
And after add:
```php
$this->oldAccessList['users'] = $this->getAffectedUsers($this->oldAccessList['users'], $this->oldParentId);
```

nickvergessen: > Because the new users will be notified (included in the previous line) and the old users, that loose the visibility to the file, don't need to be no notified. Correct?

No, they should be notified that the file is deleted.

chrisfritsche: @nickvergessen Great!! I got it now, make the adjusts and tested it! Perfect! Thx a lot!

chrisfritsche:
Solve [issue](https://github.com/nextcloud/groupfolders/issues/1018).

nickvergessen:
Fine by me. I'm time limited and won't be able to do it myself.

fschrempf:
Here is a follow-up PR: #693.

PR #506, opened 2020-10-05
getUsersFavoritingObject is moved to the manager

nickvergessen:
Fix #499 

- [x] Requires: https://github.com/nextcloud/server/pull/23184
Ref https://sentry.io/organizations/nextcloud-gmbh/issues/1927876255

PR #508, opened 2020-10-05
Fix the deletion error in public links

sharidas:
When user tries to delete the files in a public
link, the files don't get deleted. The tagManager
gets null and there was no check to handle this.
Added check for this.

Signed-off-by: Sujith Haridasan <sujith.h@gmail.com>

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -190,16 +190,20 @@ class FilesHooks {
 		$filteredEmailUsers = $this->userSettings->filterUsersBySetting($users, 'email', Files::TYPE_FILE_CHANGED);
 		$filteredNotificationUsers = $this->userSettings->filterUsersBySetting($users, 'notification', Files::TYPE_FILE_CHANGED);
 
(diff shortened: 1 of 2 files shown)


sharidas:
### Summary

When user tries to delete the files inside a public link the files are not getting deleted. The permission set for the public link in this case is `Allow upload and editing`.

### Description

The tagManager is set to null in the stack trace. Hence I have tried to provide an array with nulls if the tagManager is null, which aligns with the function signature( function is `getFileChangeActivitySettings`).

### Steps done for testing

1. Create a public link with folder. Add some files to this folder.
2. Give the permission to this link as `Allow upload and editing`
3. Open this link in an incognito mode
4. Delete the file from this link.
5. No error found in the logs or in the networking tab.
6. The file was found in the trash.

nickvergessen:
See #506

nickvergessen:
Thanks for your PR anyway! 😎
Always cool to see PRs from people with OC t-shirts on their avatars 👋

PR #509, opened 2020-10-06
Fix mail table columns to match the activity table

### LINKED ISSUE #503: amq_subjectparams has a different size than subjectparams
The subjectparams table of oc_activity_mq has a size limit of 4000 characters while the subjectparams of oc_activity are a longtext column. 

@nickvergessen Is there any reasoning behind that? Wouldn't it make sense to have a longtext for the mq table as well, since otherwise activity entries might not be added to the mq table if the subject params json exceeds 4000 chars.

nickvergessen:
Fix #503

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -14,7 +14,7 @@
 		More information is available in the Activity documentation.
 	</description>
 
-	<version>2.14.0</version>
+	<version>2.14.1</version>
 	<licence>agpl</licence>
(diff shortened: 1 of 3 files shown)


juliushaertl (approved):

juliushaertl:
Tests not happy with the app id shortening

nickvergessen:
the string was 33 chars...

PR #510, opened 2020-10-06
Fix the comments added to the activity table

sharidas:
When user adds comments to a file, the activity
notifications were not updated. This change does
fix the issue.

Signed-off-by: Sujith Haridasan <sujith.h@gmail.com>

### DIFF
--- a/lib/Consumer.php
+++ b/lib/Consumer.php
@@ -67,10 +67,7 @@ class Consumer implements IConsumer {
 		$emailSetting = $this->userSettings->getUserSetting($event->getAffectedUser(), 'email', $event->getType());
 		$emailSetting = ($emailSetting) ? $this->userSettings->getUserSetting($event->getAffectedUser(), 'setting', 'batchtime') : false;
 
-		// Add activity to stream
-		if (!$selfAction) {
-			$activityId = $this->data->send($event);
-		}
+		$activityId = $this->data->send($event);
 
 		if (!$selfAction && $notificationSetting && $activityId) {
 			$this->notificationGenerator->sendNotificationForEvent($event, $activityId);
--- a/tests/ConsumerTest.php
+++ b/tests/ConsumerTest.php
@@ -144,7 +144,7 @@ class ConsumerTest extends TestCase {
 	 * @param string $subject
 	 * @param array|false $expected
 	 */
-	public function testReceiveStream(string $type, string $author, string $affectedUser, string $subject, $expected): void {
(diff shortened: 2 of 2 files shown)


sharidas:
### Summary
The comments added for files were missing in the activity app.

### Description
The comments added for the files were working in stable19 branch but were found to be not working in stable20 branch. This change helps to fix the issue.

### How was tested?
- Enable activity app
- Add a file to the Filesystem
- Add a comment to the file
- Navigate to the activity tab of the file, the comment is seen, just like stable19.

lib/Consumer.php
```diff
@@ -68,11 +68,12 @@ public function receive(IEvent $event) {
 		$emailSetting = ($emailSetting) ? $this->userSettings->getUserSetting($event->getAffectedUser(), 'setting', 'batchtime') : false;
 
 		// Add activity to stream
-		if (!$selfAction) {
+		if ($selfAction) {
 			$activityId = $this->data->send($event);
 		}
```
nickvergessen, line 73:
```suggestion
		$activityId = $this->data->send($event);
```

lib/Consumer.php
```diff
@@ -68,11 +68,12 @@ public function receive(IEvent $event) {
...
 
-		if (!$selfAction && $notificationSetting && $activityId) {
+		//Check if the activityId is set before using the variable directly in the condition.
+		if ($selfAction && $notificationSetting && isset($activityId)  && $activityId) {
```
nickvergessen, line 76:
```suggestion
		if (!$selfAction && $notificationSetting && $activityId) {
```

nickvergessen (changes requested):
This is how it should be.
Currently your own comments are missing, but with your current change the comments of others would be missing.

lib/Consumer.php
sharidas: Done

lib/Consumer.php
sharidas: Now I have verified with the scenario:

1. Created users `admin`, `user1`.
2. `admin` shares a file `foo.txt` to `user1`
3. `admin` comments on `foo.txt`. `admin` can see the comments and so does `user1`
4. `user1` comments on `foo.txt`. Both `admin` and `user1` can see the comments.

5. `admin` creates a file `bar.txt`
6. `admin` comments on the file`bar.txt` (self). The activity is available in the table.

nickvergessen (approved):

PR #546, opened 2021-01-14
Don't break sending emails with non-string parameters

nickvergessen:
A workaround for https://github.com/nextcloud/server/issues/24434
The problem is int is invalid for names of rich object strings and might actually break the mobile clients on parsing them.
So finding out which app it actually is and turn it to a string there is the better way.


artonge (approved):

artonge:
Rebased

PR #550, opened 2021-01-21
Fix doctrine compatibility

nickvergessen:

### DIFF
--- a/tests/DataDeleteActivitiesTest.php
+++ b/tests/DataDeleteActivitiesTest.php
@@ -23,12 +23,12 @@ declare(strict_types=1);
 
 namespace OCA\Activity\Tests;
 
-use Doctrine\DBAL\Driver\Statement;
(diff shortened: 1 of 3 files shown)


tests/DataDeleteActivitiesTest.php
```diff
@@ -121,7 +121,7 @@ protected function assertUserActivities(array $expected) {
 		$this->assertTableKeys($expected, $query, 'affecteduser');
 	}
 
-	protected function assertTableKeys(array $expected, Statement $query, string $keyName): void {
+	protected function assertTableKeys(array $expected, PreparedStatement $query, string $keyName): void {
```
ChristophWurst, line 124: why not the interface?

nickvergessen: right

ChristophWurst (approved):
Smooth

PR #565, opened 2021-03-18
Fix notifying own activities

tcitworld:
`$createStream` was unused and the `|| true` was odd. This *should* be correct behaviour as far as I understand it.

### DIFF
--- a/lib/NotificationGenerator.php
+++ b/lib/NotificationGenerator.php
@@ -49,10 +49,7 @@ class NotificationGenerator implements INotifier {
 		$selfAction = $event->getAffectedUser() === $event->getAuthor();
 		$notifySetting = $this->userSettings->getUserSetting($event->getAffectedUser(), 'notify', $event->getType());
 
-		// User is not the author or wants to see their own actions
-		$createStream = !$selfAction || $this->userSettings->getUserSetting($event->getAffectedUser(), 'setting', 'self');
-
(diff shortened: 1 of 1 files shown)


lib/NotificationGenerator.php
```diff
@@ -52,7 +52,7 @@ public function sendNotificationForEvent(IEvent $event, int $activityId) {
 		// User is not the author or wants to see their own actions
 		$createStream = !$selfAction || $this->userSettings->getUserSetting($event->getAffectedUser(), 'setting', 'self');
 
-		if (($notifySetting || true) && $event->getGenerateNotification()) {
+		if ($notifySetting && $createStream && $event->getGenerateNotification()) {
```
nickvergessen, line 55:
```suggestion
		if ($notifySetting && !$selfAction && $event->getGenerateNotification()) {
```

This line which is above can be removed (github diff just doesn't allow me to post that):
```php
$createStream = !$selfAction || $this->userSettings->getUserSetting($event->getAffectedUser(), 'setting', 'self');
```
The notification has nothing to do with `$createStream`

Also `'setting', 'self'` has been removed.

Actually what you would want is:

tcitworld: > Also 'setting', 'self' has been removed.

Thanks, I missed that.

icewind1991 (approved):
:see_no_evil:

nickvergessen (approved):

PR #568, opened 2021-03-25
Add dedicated activities for file uploads to public shares

MorrisJobke:
The events are handled in the files app: https://github.com/nextcloud/server/pull/26300

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -156,10 +156,12 @@ class FilesHooks {
 			return;
 		}
 
+		$isDir = $this->view->is_dir($path);
+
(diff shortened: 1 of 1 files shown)


nickvergessen:
But maybe I didn't upload but only create it on the instance with the existing options in the + menu?

MorrisJobke:
> But maybe I didn't upload but only create it on the instance with the existing options in the + menu?

Which is still fine to be handled as "Uploaded a template"?

nickvergessen:
Another thing to consider, it does not group anymore.
So when you have a sync client and dump in a folder with 3 subfolder and 3 files each, you get 6 activity rows instead of 1?

PR #576, opened 2021-04-13
Ensure link names are unique for accessibility, thanks @nickvergessen, fix #575

### LINKED ISSUE #575: Duplicate links on page with generic title "Open file"
Currently there are a lot of duplicate links on the page all saying "Open file", which makes it a problem for screen readers:
image

We have 2 ways of fixing this:
- We could change the text to "Open `filename`" stating the actual filename. But actually the links are already there in text form directly above. So I would rather propose:
- We change the previews into div elements rather than links so they are not picked up by screenreaders and not read out duplicate. We can still attach JS event to them so on click the file gets opened, but then if you keyboard-tab through the page you won’t get hit by duplicates.

jancborchardt:
Before, there were a lot of duplicate links on the page all saying "Open file", which makes it a problem for screen readers:
image

This is now fixed by changing the link to say "Open `filename`" stating the actual filename.

Since the links are already there in text form directly above, in the future we could also remove the duplicate linked images from screenreader flow by using divs for them and triggering the link via Javascript.

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -253,7 +253,7 @@ $(function(){
 				for (var i = 0; i < activity.previews.length; i++) {
 					var preview = activity.previews[i];
 					content += ((preview.link) ? '<a href="' + preview.link + '">' + "\n" : '')
-						+ '<img class="preview' + ((preview.isMimeTypeIcon) ? ' preview-mimetype-icon' : '') + '" src="' + preview.source + '" alt="' + t('activity', 'Open file') + '" />' + "\n"
+						+ '<img class="preview' + ((preview.isMimeTypeIcon) ? ' preview-mimetype-icon' : '') + '" src="' + preview.source + '" alt="' + t('activity', 'Open {filename}', preview) + '" />' + "\n"
 						+ ((preview.link) ? '</a>' + "\n" : '')
 				}
 				content += '</div>';
--- a/lib/Controller/APIv2Controller.php
+++ b/lib/Controller/APIv2Controller.php
@@ -402,6 +402,7 @@ class APIv2Controller extends OCSController {
 			'isMimeTypeIcon' => true,
 			'fileId' => $fileId,
 			'view' => $info['view'] ?: 'files',
+			'filename' => basename($filePath),
 		];
 	}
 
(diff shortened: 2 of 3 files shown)


js/script.js
```diff
@@ -253,7 +253,7 @@ $(function(){
 				for (var i = 0; i < activity.previews.length; i++) {
 					var preview = activity.previews[i];
 					content += ((preview.link) ? '<a href="' + preview.link + '">' + "\n" : '')
-						+ '<img class="preview' + ((preview.isMimeTypeIcon) ? ' preview-mimetype-icon' : '') + '" src="' + preview.source + '" alt="' + t('activity', 'Open file') + '" />' + "\n"
+						+ '<img class="preview' + ((preview.isMimeTypeIcon) ? ' preview-mimetype-icon' : '') + '" src="' + preview.source + '" alt="' + t('activity', 'Open ') + preview.filename + '" />' + "\n"
```
nickvergessen, line 256:
```suggestion
						+ '<img class="preview' + ((preview.isMimeTypeIcon) ? ' preview-mimetype-icon' : '') + '" src="' + preview.source + '" alt="' + t('activity', 'Open {filename}', preview) + preview.filename + '" />' + "\n"
```

nickvergessen (changes requested):

nickvergessen (approved):

nickvergessen:
Pushed a fix to the unit tests

lib/Controller/APIv2Controller.php
```diff
@@ -363,6 +363,7 @@ protected function getPreview(string $owner, int $fileId, string $filePath): arr
 			'isMimeTypeIcon' => true,
 			'fileId' => $fileId,
 			'view' => $info['view'] ?: 'files',
+			'filename' => basename($filePath),
```
MorrisJobke, line 366: @nickvergessen This is an REST API change, shouldn't it be documented?

nickvergessen: @jancborchardt can you add it in https://github.com/nextcloud/activity/blob/master/docs/endpoint-v2.md#preview-element ? :P

MorrisJobke: https://github.com/nextcloud/activity/pull/588

PR #581, opened 2021-04-22
Rewrite the sidebar tab in Vue

artonge:
This is a rewrite in Vue of the sidebar tab.

I used [`@juliushaertl/vue-richtext`](https://github.com/juliushaertl/vue-richtext) to render activities, and created components for activity types that needed it.


package.json
```diff
@@ -0,0 +1,105 @@
+{
+	"name": "activity",
+	"description": "This application enables users to view actions related to their files in Nextcloud.",
+	"version": "2.15.0",
+	"author": "Julius Härtl <jus@bitgrid.net",
```
skjnldsv, line 5: A changer

src/components/Activity.vue
```diff
@@ -0,0 +1,198 @@
...
+<style lang="scss" scoped>
+.activity-entry {
+	display: flex;
+	align-items: flex-start;
+	width: 100%;
+	height: var(--height);
+	min-height: 32px;
+	padding-bottom: 5px;
+
+	.activity-entry-icon {
```
skjnldsv, line 173:
```suggestion
	.activity-entry__icon {
```

src/components/Activity.vue
```diff
@@ -0,0 +1,198 @@
...
+		opacity: 0.5;
+		margin-right: 8px;
+	}
+
+	.activity-entry-content {
```
skjnldsv, line 178:
```suggestion
	&__content {
```

src/components/richArgumentsTypes/FileRichArgument.vue
```diff
@@ -0,0 +1,84 @@
...
+	a {
+		font-weight: bold;
+
+		&:hover {
+			opacity: 0.7;
+			text-decoration: underline;
+		}
+	}
+
+	.icon-home {
```
skjnldsv, line 79:
```suggestion
	::v-deep .icon-home {
```

src/components/richArgumentsTypes/CalendarEventRichArgument.vue
```diff
@@ -0,0 +1,49 @@
...
+<style lang="scss" scoped>
+	a {
+		font-weight: bold;
+
+		&:hover {
+			opacity: 0.7;
+			text-decoration: underline;
+		}
+	}
+</style>
```
skjnldsv, line 49: Please put this in the Activity component with a v-deep :)

src/models/ActivityModel.js
```diff
@@ -0,0 +1,279 @@
...
+	 * inject its watchers into the _activity
+	 * state and make the whole class reactive
+	 *
+	 * @returns {Object} the activity raw state
+	 * @readonly
+	 * @memberof Sidebar
+	 */
+	get state() {
+		return this._activity
+	}
```
skjnldsv, line 60: Should not be necessary since we switched from `#activity` to `_activity` which vue likes ;)

src/views/ActivityTab.vue
```diff
@@ -0,0 +1,146 @@
...
+
+<template>
+	<div :class="{ 'icon-loading': loading }">
+		<!-- error message -->
+		<div
+			v-if="error"
+			class="emptycontent">
+			<div class="icon icon-error" />
+			<h2>{{ error }}</h2>
+		</div>
```
skjnldsv, line 31: Use EmptyContent :)

cypress/cypress.json
```diff
@@ -0,0 +1 @@
+{}
```
skjnldsv, line 1: To delete?

src/views/ActivityTab.vue
```diff
@@ -0,0 +1,148 @@
...
+			 */
+			loading: true,
+			/**
+			 * @type {Object}
+			 */
+			fileInfo: null,
+			/**
+			 * @type {Array.<Object>}
+			 */
+			activities: [],
```
skjnldsv, line 77: We don't use types in data, I'm not convinced it really adds anything?
Most IDE already understand the default value in vue I think?

src/models/ActivityModel.js
```diff
@@ -0,0 +1,264 @@
...
+
+	/**
+	 * Create the activity object
+	 *
+	 * @param {Object} ocsData ocs request response
+	 */
+	constructor(ocsData) {
+		if (ocsData.ocs && ocsData.ocs.data && ocsData.ocs.data[0]) {
+			ocsData = ocsData.ocs.data[0]
+		}
```
skjnldsv, line 37: I would just check if that's an object and throw if not.

Then add proper basic mandatory fields checks like `id`.

Having fallback and data check on constructor feels out of place, no? Either it's called properly or it throw? Safety data changes to support multiple parameters seems wrong to me :thinking:

src/components/richArgumentsTypes/OpenGraphRichArgument.vue
```diff
@@ -0,0 +1,50 @@
...
+<script>
+export default {
+	name: 'OpenGraphRichArgument',
+	props: {
+		id: { type: String, required: true },
+		name: { type: String, required: true },
+		description: { type: String, required: true },
+		thumb: { type: String, default: '' },
+		link: { type: String, required: true },
+		website: { type: String, required: true },
```
skjnldsv, line 42: Please expand the objects

src/components/richArgumentsTypes/FileRichArgument.vue
```diff
@@ -0,0 +1,77 @@
...
+
+export default {
+	name: 'FileRichArgument',
+	directives: {
+		tooltip: Tooltip,
+	},
+	props: {
+		name: { type: String, required: true },
+		path: { type: String, required: true },
+		link: { type: String, required: true },
```
skjnldsv, line 42: Please expand the objects

src/components/richArgumentsTypes/EmailRichArgument.vue
```diff
@@ -0,0 +1,34 @@
...
+<template>
+	<a :href="`mailto:${id}`">{{ name }}</a>
+</template>
+
+<script>
+export default {
+	name: 'EmailRichArgument',
+	props: {
+		id: { type: String, required: true },
+		name: { type: String, required: true },
```
skjnldsv, line 31: Please expand the objects

src/components/richArgumentsTypes/CalendarEventRichArgument.vue
```diff
@@ -0,0 +1,39 @@
...
+</template>
+
+<script>
+
+export default {
+	name: 'CalendarEventRichArgument',
+	props: {
+		name: { type: String, required: true },
+		title: { type: String, required: true },
+		link: { type: String, required: true },
```
skjnldsv, line 36: Please expand the objects

skjnldsv (changes requested):
See comments :)
Looks great!

Let's ping Jan for design update, now this is the time to do that :rocket:

jancborchardt:
Nice nice @artonge! :) 2 things from the design side:
- [ ] The icons are a bit larger than previously, and as that also larger than the tab icons. Is this intentional? Smaller looks a bit nicer
- [ ] The whitespace between entries was reduced – but we should have some more breathing room, actually more than before even. We try to use units on a scale of 4, so you can check if e.g. 8 or 12 or 16px of padding looks nicer.
- [ ] This might not be something for this pure rewrite – but it would be quite nice if for the "[person] commented" activity instead of an icon, we use their avatar.

artonge:
**UI update:**

> - The icons are a bit larger than previously, and as that also larger than the tab icons. Is this intentional? Smaller looks a bit nicer

I felt the opposite, I felt that they were to small previously, they felt "cheap" to me compared to the *Sharing* and *Comments* tab. Maybe we could match those ?

> - The whitespace between entries was reduced – but we should have some more breathing room, actually more than before even. We try to use units on a scale of 4, so you can check if e.g. 8 or 12 or 16px of padding looks nicer.

Screenshot from 2021-05-06 12-46-44

src/components/Activity.vue
```diff
@@ -0,0 +1,240 @@
...
+	<li class="activity-entry">
+		<Avatar
+			v-if="activity.type !== 'comments'"
+			class="activity-entry__icon"
+			:disable-menu="true"
+			:disable-tooltip="true"
+			:url="activity.icon"
+			:size="15" />
+		<Avatar
+			v-if="activity.type === 'comments'"
```
nickvergessen, line 32:
```suggestion
			v-else
```

src/main.js
```diff
@@ -0,0 +1,65 @@
+/**
```
nickvergessen, line 1: Might be a bit confusing if main is not the page of the app but the sidebar?
Maybe call this sidebar and leave main free to the "main page" which is apps/activity ?

src/views/ActivityTab.vue
```diff
@@ -0,0 +1,136 @@
...
+
+<template>
+	<div :class="{ 'icon-loading': loading }">
+		<!-- error message -->
+		<EmptyContent v-if="error" icon="icon-error">
+			{{ error }}
+		</EmptyContent>
+
+		<!-- activities content -->
+		<ul>
```
nickvergessen, line 31: v-else? or what happens when you managed to load a chunk of activities and then get an error so the empty content from above shows?

nickvergessen (approved):
Woho, I made 3 attempts to migrate the normal activity view to Vue, but always failed to finish it completely.
This looks like a step into the future :)

nickvergessen:
> This might not be something for this pure rewrite – but it would be quite nice if for the "[person] commented" activity instead of an icon, we use their avatar.

Please let's not do this here. This needs fixing on various levels and should be reflected via the API so it works in the mobile apps then too.

rullzer:
Looking super slick @artonge 
I vote to try to get it in asap. So that we can work on smaller fixes separatly. As to not make this bigger than it has to be right now.

CC: @skjnldsv

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,76 @@
...
+        with:
+          repository: nextcloud/server
+          ref: ${{ matrix.server-versions }}
+
+      - name: Checkout submodules
+        shell: bash
+        run: |
+          auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+          git submodule sync --recursive
+          git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
```
skjnldsv, line 35: What is this btw?

skjnldsv (approved):

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,70 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [12.x]
+        # containers: [1, 2, 3]
+        php-versions: [ '7.4' ]
+        databases: [ 'sqlite' ]
+        server-versions: [ 'master' ]
+
```
skjnldsv, line 22:
```suggestion

    name: cypress ${{ matrix.node-version }}
```

PR #583, opened 2021-05-04
Use PNG images in daily activity summary emails

eneiluj:
This way we stay consistent with notification emails.


nickvergessen (approved):

eneiluj:
@nickvergessen Thanks for the review. What do you think about backporting this to stable20 and stable21?

nickvergessen:
Sure

PR #588, opened 2021-05-12
Add filename in the preview docs

MorrisJobke:
cc @jancborchardt @tobiasKaminsky @nickvergessen 

For https://github.com/nextcloud/activity/pull/576/files#r630185430

### DIFF
--- a/docs/endpoint-v2.md
+++ b/docs/endpoint-v2.md
@@ -102,6 +102,7 @@ Field name | Type | Value description
 `fileId` | int | The if of the actual file
 `view` | string | The view where the file can be found (either `files` or `trashbin`)
 `isMimeTypeIcon` | bool | True if `source` points to a mime type icon instead of a real preview
+`filename` | string | The filename of the file the preview is for
 
 In case the endpoint returns more fields, they should be ignored and are deprecated (only for backwards compatibility usage) or internal.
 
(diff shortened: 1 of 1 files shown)


MorrisJobke:
I just wonder why this is added to the preview and not to the element itself 🤔

MorrisJobke:
ref #576

MorrisJobke:
@nickvergessen Tobias wondered yesterday if this should be seen as breaking change? There are JSON parsers out there that also break if new elements are added.

nickvergessen:
> nickvergessen Tobias wondered yesterday if this should be seen as breaking change? There are JSON parsers out there that also break if new elements are added.

Adding additional items is per semver not a breaking change, so it should be fine. No need to duplicate the API unnecessarily.
Also any app could create a new object type and the clients should gracefully survive this.

> I just wonder why this is added to the preview and not to the element itself 🤔

Because the preview object is disconnected from the files object (not every file can has a preview) and it was just easier this way to inject the data on the level it is needed and available instead of starting to map values later on.

nickvergessen (approved):

tobiasKaminsky:
> Adding additional items is per semver not a breaking change, so it should be fine. No need to duplicate the API unnecessarily.
> Also any app could create a new object type and the clients should gracefully survive this.

ok, did not knew this. I already checked with other clients to see that everything is working.

PR #593, opened 2021-06-14
Use npm 7

artonge:
:warning:  Required checks need to be updated.


.eslintrc.js
```diff
@@ -1,5 +1,12 @@
 module.exports = {
 	extends: [
 		'@nextcloud',
-	]
+	],
+	rules: {
+		// Allow shallow import of @vue/test-utils in order to be able to use it in
+		// the src folder
```
skjnldsv, line 7: Why is this needed in the src folder? :thinking:

skjnldsv: Ah, src/tests?
You can create another eslintrc there then, that should extend or supersed this config :wink:

package.json
```diff
@@ -28,7 +28,7 @@
 		"test:update-snapshots": "NODE_ENV=test jest --updateSnapshot"
 	},
 	"engines": {
-		"node": ">=10.0.0"
+		"node": ">=14.0.0"
```
skjnldsv, line 31:
```suggestion
		"node": ">=14.0.0",
		"npm": ">=7.0.0"
```

skjnldsv (approved):
Small nitpicks, otherwise good! :)

.eslintrc.js
artonge: My plan was to make a PR to add it to the shared eslint config as it already contains an exception for `@vue/test-utils`.
Is it the same use case ? https://github.com/nextcloud/eslint-config/blob/master/index.js#L97

nickvergessen (approved):
Seems to work

.eslintrc.js
skjnldsv: :scream:

skjnldsv: https://github.com/nextcloud/eslint-config/pull/26 .........

artonge: Not sure if that's a yes or a no :thinking:

skjnldsv: I did not see that pass by, I'm not sure we should have done that in the first place, but yeah :shrug:

Issue #582, opened 2021-04-28
"by you" tab doesn't show created or deleted ,files and folders by you


BabiWD:
Hello I am new to GitHub so if I am not posting this issue or question the right way I sincerely apologize please guide me the right way.
So my problem with Activity app is that in the left menu there is a tab called "by you". I think it should show all of your activities  but it doesn't, it does  editing but it doesn't show created and deleted files!
You might ask why do I need such log, well that's because I use impersonate app to check users activity and it's fine when you are in "all activities" but it's annoying to also see others activities as well and when you switch to "by you" tab it's not showing created/deleted folders and files for some reason and I couldn't find an proper answer online so I thought I should be asking this here if it's a bug or a feature that is for privacy or something that I don't know!
if there is anyone experienced please help me out and thanks.

jakobroehrl:
Maybe similar to: https://github.com/nextcloud/activity/issues/514

jakobroehrl:
I did not see any differences in the table:


MariaDB [nc23]> select * from oc_activity where user='root' and type='file_created' or type='file_deleted' or type='file_changed';
+-------------+------------+----------+--------------+-------+--------------+-------+--------------+--------------------------------------------------------+---------+---------------+----------------------------------+-------------------------------------------------------------------+-------------+-----------+
| activity_id | timestamp  | priority | type         | user  | affecteduser | app   | subject      | subjectparams                                          | message | messageparams | file                             | link                                                              | object_type | object_id |
+-------------+------------+----------+--------------+-------+--------------+-------+--------------+--------------------------------------------------------+---------+---------------+----------------------------------+-------------------------------------------------------------------+-------------+-----------+
|          64 | 1625561540 |       30 | file_created | root  | root         | files | created_self | [{"296":"\/new.md"}]                                   |         | []            | /new.md                          | http://localhost/server23/index.php/apps/files/?dir=/             | files       |       296 |
|          65 | 1625561551 |       30 | file_changed | root  | root         | files | changed_self | [{"296":"\/new.md"}]                                   |         | []            | /new.md                          | http://localhost/server23/index.php/apps/files/?dir=/             | files       |       296 |
|          66 | 1625561571 |       30 | file_deleted | root  | root         | files | deleted_self | [{"178":"\/document.txt"}]                             |         | []            | /document.txt                    | http://localhost/server23/index.php/apps/files/?dir=/             | files       |       178 |
[...]

PR #610, opened 2021-07-07
show add, del and restored files within by and self filter

jakobroehrl:
Signed-off-by: Jakob Röhrl <jakob.roehrl@web.de>
Fixes: #582

### DIFF
--- a/lib/UserSettings.php
+++ b/lib/UserSettings.php
@@ -154,9 +154,14 @@ class UserSettings {
 	 */
 	public function getNotificationTypes() {
 		$settings = $this->manager->getSettings();
-		return array_map(function (ActivitySettings $setting) {
+		
+		$return = array_map(function (ActivitySettings $setting) {
 			return $setting->getIdentifier();
 		}, $settings);
+		if (array_search('file_changed', $return) !== false) {
+			array_push($return, 'file_created', 'file_deleted', 'file_restored');
+		}
+		return $return;
 	}
 
 	/**


lib/UserSettings.php
```diff
@@ -154,9 +154,13 @@ protected function getDefaultFromSetting($method, $type) {
...
 	public function getNotificationTypes() {
 		$settings = $this->manager->getSettings();
-		return array_map(function (ActivitySettings $setting) {
+		
+		$ret = array();
+		$ret = array_map(function (ActivitySettings $setting) {
 			return $setting->getIdentifier();
 		}, $settings);
+		array_push($ret, 'file_created', 'file_deleted', 'file_restored');
+		return $ret;
```
nickvergessen, line 163:
```suggestion
		$return = array_map(function (ActivitySettings $setting) {
			return $setting->getIdentifier();
		}, $settings);
		array_push($return, 'file_created', 'file_deleted', 'file_restored');
		return $return;
```

Also check for `file_changed` before adding them with https://www.php.net/manual/en/function.array-search.php

jakobroehrl: Thanks, fixed it!

lib/UserSettings.php
```diff
@@ -154,9 +154,14 @@ protected function getDefaultFromSetting($method, $type) {
 	 */
 	public function getNotificationTypes() {
 		$settings = $this->manager->getSettings();
-		return array_map(function (ActivitySettings $setting) {
+		
+		$return = array_map(function (ActivitySettings $setting) {
 			return $setting->getIdentifier();
 		}, $settings);
+		if (array_search('file_changed', $return)) {
```
nickvergessen, line 161:
```suggestion
		if (array_search('file_changed', $return) !== false) {
```

otherwise it doesn't work if `file_changed` is in place `0`

jakobroehrl:
Thanks, let's merge it! :)

nickvergessen (approved):

PR #617, opened 2021-07-09
Only call getCloudId if needed

juliushaertl:
[getCloudId can be expensive](https://github.com/nextcloud/server/blob/master/lib/private/Federation/CloudIdManager.php#L84-L120), especially on large instances since curently the contacts manager is called to search for the cloud id which may lead to slow queries as it searches the oc_card_properties table with a LIKE on every file action:

SELECT DISTINCT `cp`.`cardid` FROM `oc_cards_properties` `cp` WHERE (`cp`.`addressbookid` = N) AND (`cp`.`name` = N) AND (`cp`.`value`  COLLATE utfNmbN_general_ci LIKE N);

This PR ensures that we only call this if really needed for dispatching remote activities. A separate enhancement to avoid this search for local users is following for the server.

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -229,7 +229,9 @@ class FilesHooks {
 
 		$accessList = $this->getUserPathsFromPath($filePath, $uidOwner);
 
-		$this->generateRemoteActivity($accessList['remotes'], $activityType, time(), $this->currentUser->getCloudId(), $accessList['ownerPath']);
+		if (!empty($accessList['remotes'])) {
+			$this->generateRemoteActivity($accessList['remotes'], $activityType, time(), $this->currentUser->getCloudId(), $accessList['ownerPath']);
+		}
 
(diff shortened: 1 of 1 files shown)


lib/FilesHooks.php
```diff
@@ -229,7 +229,9 @@ protected function addNotificationsForFileAction($filePath, $activityType, $subj
 
 		$accessList = $this->getUserPathsFromPath($filePath, $uidOwner);
 
-		$this->generateRemoteActivity($accessList['remotes'], $activityType, time(), $this->currentUser->getCloudId(), $accessList['ownerPath']);
+		if (!empty($accessList)) {
```
nickvergessen, line 232:
```suggestion
		if (!empty($accessList['remotes'])) {
```

?

juliushaertl: Arg, my bad, thanks for catching :see_no_evil:

skjnldsv (approved):

nickvergessen (approved):

PR #629, opened 2021-07-29
fix activity design

szaimen:
| Before | After |
|---|---|
| image | image |

Signed-off-by: szaimen <szaimen@e.mail.de>

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -10,6 +10,7 @@
 	margin-bottom: 5px;
 	margin-left: 5px;
 	text-overflow: ellipsis;
+	overflow-wrap: anywhere;
 }
 
 .activitymessage {


css/style.scss
```diff
@@ -10,6 +10,7 @@
 	margin-bottom: 5px;
 	margin-left: 5px;
 	text-overflow: ellipsis;
+	word-break: break-word;
```
Pytal, line 13:
```suggestion
	word-break: normal;
	overflow-wrap: anywhere;
```
Seems `word-break: break-word;` is deprecated https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#values

Pytal (changes requested):

css/style.scss
szaimen: done

Pytal (approved):

szaimen:
The fixup bot is still broken unfortunatley. Does anyone has force-merge privileges?
Maybe @nickvergessen?

szaimen:
@skjnldsv do you have force-merge rights here?

szaimen:
Thanks @nickvergessen ! :)

PR #609, opened 2021-07-07
Fix "Enable notification emails"

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

PR #657, opened 2021-11-19
Fix reenabling the email option

nickvergessen:
Fix #609 

### Steps
1. Go to admin settings
2. Disable the notification email option
3. Reload the page and check if the state is consistent
4. Enable the notification email option again
5. Reload the page and check if the state is consistent


PVince81 (approved):
classic jquery :+1:

PR #671, opened 2021-12-04
🧹 Move UserDeleted and PostLogin hooks to listeners

tcitworld:
Tests have been adjusted and improved as well.


tests/HooksDeleteUserTest.php
```diff
@@ -47,79 +70,115 @@ protected function setUp(): void {
...
+			->where($qb->expr()->eq('amq_type', $qb->createNamedParameter('test')));
 		$this->assertTableKeys($expected, $query, 'amq_affecteduser');
 	}
 
-	protected function assertTableKeys(array $expected, IPreparedStatement $query, string $keyName): void {
-		$query->execute();
+	/**
+	 * @throws Exception
+	 */
+	protected function assertTableKeys(array $expected, IQueryBuilder $query, string $keyName): void {
```
tcitworld, line 134: Wanted to remove this test file as well but not sure what's being asserted in this function.

nickvergessen (approved):

tests/HooksDeleteUserTest.php
nickvergessen: Whether entries are actually deleted. I would say your other tests cover this sufficient.

tcitworld: I'll remove the file then.

skjnldsv (approved):

PR #690, opened 2022-01-03
Create command-compile.yml

nickvergessen:


nickvergessen:
So we can recompile in comments like https://github.com/nextcloud/activity/pull/680#issuecomment-1004171766

artonge (approved):

PR #693, opened 2022-01-04
Fix group move/rename activities

fschrempf:
This is a respin of #500. The code is essentially the same, but with a few small style fixes and rebased and tested on master of server, activities app and groupfolders app.

This still doesn't cover the handling of sharing events for other groupfolder users. So it only partly solves https://github.com/nextcloud/groupfolders/issues/1018.

Cc: @chrisfritsche


artonge:
Seems covered by https://github.com/nextcloud/activity/pull/1163
Feel free to reopen if I am wrong.
And sorry for the "delay" in handling this PR.

PR #702, opened 2022-01-17
Update php lint workflows

artonge:
Php linting workflow is currently failing: https://github.com/nextcloud/activity/runs/4824472725?check_suite_focus=true

This PR updates php linting workflows as per https://github.com/nextcloud/.github/tree/master/workflow-templates


.github/workflows/lint-info-xml.yml
```diff
@@ -0,0 +1,31 @@
...
+  xml-linters:
+    runs-on: ubuntu-latest
+
+    name: info.xml lint
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+
+      - name: Download schema
+        run: wget https://raw.githubusercontent.com/nextcloud/appstore/master/nextcloudappstore/api/v1/release/info.xsd
```
nickvergessen, line 25: We need to use:
https://github.com/nextcloud/server/blob/master/resources/app-info-shipped.xsd

appinfo/info.xml
```diff
@@ -11,14 +11,13 @@
...
 		More information is available in the Activity documentation.
-	</description>
+</description>
 
 	<version>2.16.0</version>
 	<licence>agpl</licence>
 	<author>Frank Karlitschek</author>
 	<author>Joas Schilling</author>
 
-	<default_enable />
```
nickvergessen, line 21: This needs to stay, downloading the correct `.xsd` will fix it

artonge: Thanks for the pointer :+1:

nickvergessen (approved):

PR #718, opened 2022-01-28
Updating lint-info-xml.yml workflow from template

nextcloud-command:
Automated update of the lint-info-xml.yml workflow from https://github.com/nextcloud/.github


artonge:
Irrelevant as this is a shipped app.

PR #719, opened 2022-02-01
Use QueryBuilder for delete queries

nickvergessen:

### DIFF
--- a/lib/Data.php
+++ b/lib/Data.php
@@ -343,15 +343,19 @@ class Data {
 	 *    'field' => array('value', 'operator') => `field` operator 'value'
 	 */
 	public function deleteActivities($conditions) {
-		$sqlWhere = '';
-		$sqlParameters = $sqlWhereList = [];
-		foreach ($conditions as $column => $comparison) {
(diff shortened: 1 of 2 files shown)


lib/Data.php
```diff
@@ -343,33 +343,33 @@ public function expire($expireDays = 365) {
 	 *    'field' => array('value', 'operator') => `field` operator 'value'
 	 */
 	public function deleteActivities($conditions) {
-		$sqlWhere = '';
-		$sqlParameters = $sqlWhereList = [];
+		$delete = $this->connection->getQueryBuilder();
```
artonge, line 346:
```
 1) OCA\Activity\Tests\DataDeleteActivitiesTest::testChunkingDeleteNotUsedWhenNotOnMysql
Error: Call to a member function delete() on null

/home/runner/work/activity/activity/apps/activity/lib/Data.php:347
```

nickvergessen: There is no way to fix the test, so I removed it.
It was mocking the DB connection and expecting the calls to prepare, rowCount, etc.
But since we execute the tests nowadays on all DBs, we can will also see when the delete does not work anymore

artonge (approved):

PR #720, opened 2022-02-01
Try to reduce the load from writing

nickvergessen:


artonge:
If this gets merged, perhaps put a comment to explain why we save some `getQueryBuilder` as object properties.

lib/Data.php
```diff
@@ -97,10 +106,10 @@ public function send(IEvent $event): int {
 				'object_id' => $event->getObjectId(),
 				'object_name' => $event->getObjectName(),
 				'link' => $event->getLink(),
-			])
-			->execute();
+			]);
+		$this->insertActivity->executeStatement();
```
miaulalala, line 110: will this work for NC21?

artonge: Does not matter for this PR, but to remember if this is backported: https://github.com/nextcloud/activity/blob/master/appinfo/info.xml#L38

miaulalala: Ah, lovely, thanks!

miaulalala (approved):

artonge (approved):

PVince81:
backport possible ? (without composer changes)

nickvergessen:
Backport in https://github.com/nextcloud/activity/pull/731

PR #722, opened 2022-02-02
Allow specify a config prefix for another database connection

### LINKED ISSUE #721: Activity in extra database
Due to lots of writes (like #590) it might be useful for performance to move the activity table to a separate database.

- [ ] use separate DB connection / check if Doctrine can handle that
- [ ] config switches for configuring the second DB, off by default
- [ ] how to handle migrations ?

@nickvergessen

nickvergessen:
- [x] Requires server https://github.com/nextcloud/server/pull/30973
Fix #721

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -24,23 +24,37 @@
 
 namespace OCA\Activity\AppInfo;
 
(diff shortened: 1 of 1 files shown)


PVince81:
server PR is merged, please adjust CI :-)

PR #723, opened 2022-02-03
Shorten mail subject

tobiasKaminsky:
Signed-off-by: Tobias Kaminsky <tobias@kaminsky.me>

### DIFF
--- a/lib/MailQueueHandler.php
+++ b/lib/MailQueueHandler.php
@@ -382,7 +382,7 @@ class MailQueueHandler {
 			'activityEvents' => $activityEvents,
 			'skippedCount' => $skippedCount,
 		]);
-		$template->setSubject($l->t('Activity notification for %s', $this->getSenderData('name')));
+		$template->setSubject($l->t('Activity on %s', $this->getSenderData('name')));
 		$template->addHeader();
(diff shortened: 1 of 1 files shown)


lib/MailQueueHandler.php
```diff
@@ -382,7 +382,7 @@ protected function sendEmailToUser($userName, $email, $lang, $timezone, $maxTime
 			'activityEvents' => $activityEvents,
 			'skippedCount' => $skippedCount,
 		]);
-		$template->setSubject($l->t('Activity notification for %s', $this->getSenderData('name')));
+		$template->setSubject($l->t('Activity on %s', $this->getSenderData('name')));
```
artonge, line 385: So it would be something like "Activity on Louis"?
I am not sure of the context, but it sounds weird. Maybe "Activity from" or "Activity for"?

artonge (approved):

jancborchardt (approved):

lib/MailQueueHandler.php
artonge: Oh, ok then, that makes sense. @tobiasKaminsky can you confirm?

artonge (approved):

PR #729, opened 2022-02-10
Make background job time insensitive

nickvergessen:
Ref https://github.com/nextcloud/server/pull/30945

### DIFF
--- a/tests/AppInfo/ApplicationTest.php
+++ b/tests/AppInfo/ApplicationTest.php
@@ -103,7 +103,7 @@ class ApplicationTest extends TestCase {
 			[EmailNotification::class],
 			[EmailNotification::class, TimedJob::class],
 			[ExpireActivities::class,],
-			[ExpireActivities::class, TimedJob::class],
+			[ExpireActivities::class, \OCP\BackgroundJob\TimedJob::class],
 
 			// Controller
 			[ActivitiesController::class],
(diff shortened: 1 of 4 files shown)


miaulalala (approved):

tests/AppInfo/ApplicationTest.php
```diff
@@ -80,44 +80,9 @@ public function testContainerAppName(): void {
...
-			[APIv1Controller::class],
-			[APIv1Controller::class, Controller::class],
-			[APIv1Controller::class, OCSController::class],
-			[APIv2Controller::class],
-			[APIv2Controller::class, Controller::class],
-			[APIv2Controller::class, OCSController::class],
-			[FeedController::class],
-			[FeedController::class, Controller::class],
-			[SettingsController::class, SettingsController::class],
-			[SettingsController::class, Controller::class],
```
artonge, line 120: How come we don't need that anymore?

artonge (approved):

tests/AppInfo/ApplicationTest.php
nickvergessen: It uses normal DI, so should be fine.
The entries that I kept uses custom services defined in application class.
But we can also keep them. then let me fix the TimedJob one

artonge: If this is testing the DI logic, then it's unrelated and I am fine with not keeping it :)

PR #768, opened 2022-03-24
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/cypress/utils/index.js
+++ b/cypress/utils/index.js
@@ -20,6 +20,6 @@
  *
  */
 
-const randHash = () => Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 10)
+const randHash = () => Math.random().toString(36).replace(/[^a-z]+/g, '').slice(0, 10)
 
 export default { randHash }


artonge (approved):

artonge:
Not quite the same, but in this case, it works :)

nickvergessen:
@artonge feel free to merge

nickvergessen:
> cypress Expected — Waiting for status to be reported 

This is a bit suspicious thou :P

artonge:
> @artonge feel free to merge

I can't, need more rights for that :'(

CommanderRoot:
I have rebased the PR on the current master to hopefully restart the workflows and maybe make them work this time.

nickvergessen:
Can you rebase on top of https://github.com/nextcloud/activity/pull/789

artonge:
Thanks for this @CommanderRoot !

PR #821, opened 2022-06-09
Enable keyboard navigation to user label and avatar

Pytal:
### Requires

- [ ] https://github.com/nextcloud/server/pull/32774 (for keyboard activation to open user menu)

SCSS compilation tracked in https://github.com/nextcloud/activity/issues/822

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -162,6 +162,10 @@
 	vertical-align: middle;
 	display: inline-block;
 	margin: 0 5px 2px 3px;
+
+	&:focus-visible {
+		box-shadow: 0 0 0 2px var(--color-primary);
+	}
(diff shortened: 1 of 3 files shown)


js/script.js
```diff
@@ -282,7 +282,9 @@ $(function(){
 			$element.find('.avatar-name-wrapper').each(function() {
 				var element = $(this);
 				var avatar = element.find('.avatar');
+				avatar.attr('tabindex', 0)
```
artonge, line 285: Why not put it in the template? https://github.com/nextcloud/activity/blob/master/js/templates/userLocal.handlebars

artonge (approved):

PR #831, opened 2022-06-29
cancel everything if enable_mail is disabled

ArtificialOwl:
It seems that if an admin disable the 'Enable mail notifications' checkbox in Admin Settings / Activity after a user enable notifications by mail on some activities, a mail for those activities will still be generated. The user is also lock from opt-out because the checkbox is not available in his UI since admin disabled the mail notifications.

This will check the configuration 'enable_email' before running the background job that generate the mail notifications

### DIFF
--- a/lib/BackgroundJob/EmailNotification.php
+++ b/lib/BackgroundJob/EmailNotification.php
@@ -25,6 +25,7 @@ namespace OCA\Activity\BackgroundJob;
 
 use OC\BackgroundJob\TimedJob;
 use OCA\Activity\MailQueueHandler;
+use OCP\IConfig;
 
 /**
  * Class EmailNotification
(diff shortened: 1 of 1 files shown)


ArtificialOwl:
Reproduction steps:

- `admin` set `'Enable notification mails'` in `Admin Settings / Activity`,
- `user1` enable `mail` on `'Your group memberships were modified'`,
- `admin disable `'Enable notifications mails'` in `Admin Settings / Activity`,
- `user1` cannot disable `mail` on `Your group memberships were modified` as the checkbox is gone,
- add `user1` to a group (tested with `admin`) running the `./occ group:adduser admin user`
- after some time, `user1` will receive a mail about the activity.

ArtificialOwl:
Please note that [`getUserSetting()`](https://github.com/nextcloud/activity/blob/master/lib/UserSettings.php#L70) is not called during the cron execution of `EmailNotification.php`

PR #841, opened 2022-07-04
[Documentation] Clarify usage of Activity filters

aaronsegura:
Using `/:filter` returns 404.  Must use `/filter` without the colon.

    Signed-off-by: Aaron Segura <aaron@aaronsegura.com>

### DIFF
--- a/docs/endpoint-v2.md
+++ b/docs/endpoint-v2.md
@@ -24,8 +24,16 @@ GET /ocs/v2.php/apps/activity/api/v2/activity
 With type filter
 
 ```
-GET /ocs/v2.php/apps/activity/api/v2/activity/:filter
+GET /ocs/v2.php/apps/activity/api/v2/activity/{filter}
 ```
+
+Supported type filters can be obtained from:
+
+```
+GET /ocs/v2.php/apps/activity/api/v2/activity/filters
+```
+
+Additionally, filtering based on `object_type` and `object_id` (see below) is only valid when supplying `filter` as the type filter.  Filtering results based on `object_type` and `object_id` parameters is not supported when using any other type filter.
 
 ## Parameters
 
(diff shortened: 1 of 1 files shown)


docs/endpoint-v2.md
```diff
@@ -24,7 +24,7 @@ GET /ocs/v2.php/apps/activity/api/v2/activity
 With type filter
 
 ```
-GET /ocs/v2.php/apps/activity/api/v2/activity/:filter
+GET /ocs/v2.php/apps/activity/api/v2/activity/filter
```
nickvergessen, line 27: What was actually meant is that filter is a parameter:
```suggestion
GET /ocs/v2.php/apps/activity/api/v2/activity/{filter}
```
https://github.com/nextcloud/activity/blob/master/appinfo/routes.php#L28

nickvergessen (changes requested):

docs/endpoint-v2.md
aaronsegura: How would that translate into a curl command?  I'm only able to successfully filter using the syntax I originally submitted:

    $ curl -u admin:admin -H "OCS-APIRequest: true" 'http://localhost:8181/ocs/v2.php/apps/activity/api/v2/activity/filter?since=0&limit=1&object_type=files&object_id=30&format=json'
    {"ocs":{"meta":{"status":"ok","statuscode":200,"message":"OK"},"data":[{"activity_id":43,"app":"files_sharing","type":"shared","user":"admin","subject":"Shared with Test User","subject_rich":["Shared with {user}",{"file":{"type":"file","id":"30","name":"Nextcloud Manual.pdf","path":"Nextcloud Manual.pdf","link":"http:\/\/localhost:8181\/f\/30"},"user":{"type":"user","id":"testuser","name":"Test User"}}],"message":"","message_rich":["",[]],"object_type":"files","object_id":30,"object_name":"\/Nextcloud Manual.pdf","objects":{"30":"\/Nextcloud Manual.pdf"},"link":"http:\/\/localhost:8181\/apps\/files\/?dir=\/","icon":"http:\/\/localhost:8181\/core\/img\/actions\/share.svg","datetime":"2022-06-24T18:11:11+00:00"}]}}

^This returns proper results filtered for the given file ID.  Anything else I've tried either returns the latest activity without filtering or a 404.

aaronsegura: I think I have found the source of my confusion.

Is it intended that users are allowed to filter by object_id/object_type when using a type filter from the filters endpoint?  The object_id/object_type arguments are only respected when I use the literal string `filter` as the type filter.

From lib/Data.php:

```php
if ($filter === 'self') {
        $query->andWhere($query->expr()->eq('user', $query->createNamedParameter($user)));
} elseif ($filter === 'by') {
        $query->andWhere($query->expr()->neq('user', $query->createNamedParameter($user)));
} elseif ($filter === 'filter') {
        $query->andWhere($query->expr()->eq('object_type', $query->createNamedParameter($objectType)));
        $query->andWhere($query->expr()->eq('object_id', $query->createNamedParameter($objectId)));
}
```

My assumption was that object_id/object_type arguments should cause results to be filtered to the specific object_id/object_type no matter which type filter is provided.  Is that not the case?  Should I open an issue to have this corrected, or is it intended behavior?

aaronsegura: Nope, not confused any more.  I have updated my PR to reflect the intended behavior now that you've confirmed that's what was actually intended.

nickvergessen (approved):

PR #850, opened 2022-07-19
Fix contrast issue for add icon

jancborchardt:
Please check @nickvergessen @nimishavijay :)

I used `#3ea857`, a darkened hue of `var(--color-success)` cause that one actually does not have enough contrast. FYI @marcoambrosini @juliushaertl we should probably adjust that?

image


marcoambrosini:
> we should probably adjust that?

Not sure this is the solution, it would make buttons even darker. Our Nextcloud blue fails in the same way, but I don't think we should darken it, we could just avoid using it for text or for tiny important symbols.
Couldn't we avoid the green plus icon and invert it (green background and white icon)?

jancborchardt:
> Not sure this is the solution, it would make buttons even darker. Our Nextcloud blue fails in the same way, but I don't think we should darken it, we could just avoid using it for text or for tiny important symbols.

Not really true: The current color-success is not accessible in any respect, below 3:1 – so we should actually not use it anywhere. Versus the Nextcloud blue which is above 3:1:
color-success|Nextcloud blue
-|-
image|image

juliushaertl:
I'd also say darkening that to match the pass on AA would make sense. For passing AAA we could probably darken further just for the high contrast theme?

jancborchardt:
@juliushaertl yep, sounds good.

I’ll open an issue in the Vue components regarding the color variables.

juliushaertl:
server please ;)

marcoambrosini:
Oh! I thought that the contrast comparison you posted was the one of the current success color, my bad. I agree then, let's make a server pr and darken it :)

jancborchardt:
Issue opened at https://github.com/nextcloud/server/issues/33278 :)
Could we merge this then?

marcoambrosini (approved):

jancborchardt:
And sorry for the confusion @marcoambrosini! :) Should have posted a 2 comparison screenshots instead.

marcoambrosini:
btw @jancborchardt this particular icon could still fall in the "Normal text" category, so still not pass the test depending on how it's used. I'm not entirely sure about that

jancborchardt:
It's a "graphical object" (decorative one at that), and as such it passes. :)

nickvergessen (approved):

nickvergessen (approved):

jancborchardt:
@nickvergessen for dark mode it needs a different icon, same for the "deleted" icon. There we could also just use the light icons?

nickvergessen:
WCAG Contrast checker said the color was fine?

Not going to fiddle around and split the color mode based on the theme. If green does not work, we use black/white everywhere

PR #880, opened 2022-08-20
Update .l10nignore

rakekniven:
Reported at Transifex.

Signed-off-by: rakekniven <2069590+rakekniven@users.noreply.github.com>

### DIFF
--- a/.l10nignore
+++ b/.l10nignore
@@ -1,6 +1,14 @@
+js/activity-adminSettings.js
+js/activity-adminSettings.js.map 
+js/activity-adminSettings.js.LICENSE.txt
 js/activity-dashboard.js
 js/activity-dashboard.js.map
 js/activity-dashboard.js.LICENSE.txt
+js/activity-personalSettings.js
+js/activity-personalSettings.js.map
+js/activity-personalSettings.js.LICENSE.txt
+js/activity-settings.js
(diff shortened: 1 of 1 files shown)


.l10nignore
```diff
@@ -1,6 +1,14 @@
...
 js/activity-dashboard.js.map
 js/activity-dashboard.js.LICENSE.txt
+js/activity-personalSettings.js
+js/activity-personalSettings.js.map
+js/activity-personalSettings.js.LICENSE.txt
+js/activity-settings.js
+js/activity-settings.js.map
 js/activity-sidebar.js
 js/activity-sidebar.js.map
 js/activity-sidebar.js.LICENSE.txt
```
nickvergessen, line 14:
```suggestion
js/activity-*
```

nickvergessen (changes requested):

nickvergessen (approved):

nickvergessen:
Aparently it is a bit random which ignore file is used, so backporting this to all versions that still sync

nickvergessen:
Didn't work, so I checked the code, its's just a string position, so without the asterix it works, but no wildcards @rakekniven 
https://github.com/nextcloud/docker-ci/blob/master/translations/translationtool/src/translationtool.php#L200-L202

PR #899, opened 2022-08-29
Add role=navigation and aria-current

PVince81:
For https://github.com/nextcloud/server/issues/33739

### DIFF
--- a/templates/stream.app.navigation.php
+++ b/templates/stream.app.navigation.php
@@ -25,7 +25,7 @@
 <div id="app-navigation" role="navigation">
 	<ul>
 	<?php foreach ($_['navigations'] as $navigation) { ?>
-		<li<?php if ($_['activeNavigation'] === $navigation['id']): ?> class="active"<?php endif; ?>>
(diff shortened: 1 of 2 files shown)


templates/stream.app.navigation.php
```diff
@@ -22,7 +22,7 @@
  *
  */
 ?>
-<div id="app-navigation">
+<div id="app-navigation" role="navigation">
 	<ul>
 	<?php foreach ($_['navigations'] as $navigation) { ?>
 		<li<?php if ($_['activeNavigation'] === $navigation['id']): ?> class="active"<?php endif; ?>>
```
nickvergessen, line 28: Also add `aria-current=page` on the initial loaded active navigation?

```suggestion
		<li<?php if ($_['activeNavigation'] === $navigation['id']): ?> class="active" aria-current="page"<?php endif; ?>>
```

artonge (approved):

nickvergessen (approved):

nickvergessen:
Cypress unrelated

PR #954, opened 2022-10-11
Update .gitignore with more vim files

fridokus:
Add some more vim files to `.gitignore`. 

* `*.sw[a-p]` are vim swap files that could persist if vim exits ungracefully
* `[._]*.un~` are vim [undo files](https://jovicailic.org/2017/04/vim-persistent-undo/)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,8 +40,10 @@ nbproject
 .settings.xml
 .c9revisions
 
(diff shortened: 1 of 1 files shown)


artonge:
Please don't this open this kind of pull request for Oktoberfest. This adds noise, and is not helpful at all.

PR #981, opened 2022-10-26
Port sidebar icon to material

CarlSchwan:
image

Require https://github.com/nextcloud/server/pull/34769


src/sidebar.js
```diff
@@ -35,7 +37,7 @@ let ActivityTabInstance = null
 const activityTab = new OCA.Files.Sidebar.Tab({
 	id: 'activity',
 	name: t('activity', 'Activity'),
-	icon: 'icon-activity',
+	icon: LightningBolt,
```
artonge, line 40: Sexy, didn't we could do that :)

skjnldsv: DOn't get too excited, this is a bad approach I regret implementing :see_no_evil: 
We can have conflicts in vue template compilation versions between those renders methods

skjnldsv: The real good way is to import the icon as svg, similarly.
Then we can render the svg inline with sanitization. And forcing the same `color="currentColor"` trick the `vue-material-design-icons` is doing :wink: 

https://nextcloud.github.io/nextcloud-files/interfaces/Entry.html#iconSvgInline
https://github.com/nextcloud/nextcloud-upload/blob/7037cccbc34bdba250c659f300ebc9ac6d39b84a/lib/components/UploadPicker.vue#L36
https://github.com/nextcloud/nextcloud-upload/blob/7037cccbc34bdba250c659f300ebc9ac6d39b84a/lib/components/ActionIcon.vue

skjnldsv: Oh wait, I thought that was my code @CarlSchwan 
https://github.com/nextcloud/server/pull/34769/files#diff-91cd7c7f024a3f52536789f769ad017e1b60f17806d42a8342f19f67caaee4ee

I also did that in other locations :grin:

nickvergessen:
If it works, fine by me.

but cypress is unhappy

src/sidebar.js
CarlSchwan: @skjnldsv is there a good way to import material icons as svg? or do we end up copying the svg in a bunch of places?

skjnldsv: No it's as easy as what you're doing here :)
Just use the official material design icon 
https://github.com/nextcloud/nextcloud-upload/pull/444#issue-1265845849

```
import FolderPlus from '@mdi/svg/svg/folder-plus.svg'
```

artonge:
Done in https://github.com/nextcloud/activity/pull/1154

PR #992, opened 2022-10-31
apply ACL groupfolders

ArtificialOwl:
when running with `activity_use_cached_mountpoints` on a file:

- get all users [1] that have basic groupfolders' access
- get rules related to path to said file
- get list of users that might be affected by said rules
- verify access to those users
- remove users without read permission on said file from [1]


lib/FilesHooks.php
```diff
@@ -1219,4 +1210,154 @@ protected function commitActivityTransaction(bool $shouldFlush): void {
...
+				// merge current user list with members of the group
+				$usersToCheck = array_values(array_unique(array_merge($usersToCheck, $userIds)));
+			}
+		}
+
+		// now that we have a list of 'unstable' users, we confirm they have no access to the file
+		$filteredUsers = [];
+		foreach ($usersToCheck as $userId) {
+			try {
+				if (count($this->rootFolder->getUserFolder($userId)->getById($fileId)) > 0) {
```
nickvergessen, line 1352: Don't know how most groupfolder setups are and how often we end up in this code path, but if we now really set up 10k user storages when there is an negative rule, that is not going to scale and running out of memory pretty quickly.

lib/FilesHooks.php
```diff
@@ -1219,4 +1210,154 @@ protected function commitActivityTransaction(bool $shouldFlush): void {
...
+			$mapping = $rule->getUserMapping();
+			$id = $mapping->getId();
+
+			// if mapping is about user
+			if ($mapping->getType() === 'user' && !in_array($id, $usersToCheck)) {
+				$usersToCheck[] = $id;
+			}
+
+			// if mapping is about group
+			if ($mapping->getType() === 'group') {
```
nickvergessen, line 1334: Should keep a record of which groups we looped over already to avoid getting their user list more than once?

nickvergessen (changes requested):

lib/FilesHooks.php
PVince81: I don't remember how getById is setting up all mounts, it might be possible to use a shortcut and go straight to the group folder mount point name since we have it (assuming that it doesn't change in case of collisions!).
If the latter theory works, it might be possible to use `getUserfolder($userId)->get($pathToFile)` which might be more efficient.

I also wonder if we should tearDown the FS or if that's done automatically nowadays after some limit.

@icewind1991 can you advise?

nickvergessen: > I don't remember how getById is setting up all mounts, it might be possible to use a shortcut and go straight to the group folder mount point name since we have it (assuming that it doesn't change in case of collisions!).

The `getUserFolder` before will setUp the full user? Or at least load all the mount point informations (share table being contacted X times, etc)

PVince81: @nickvergessen yes and no. As far as I remember since NC 24 we have optimized the mount point setup and that part should be lazy. I hope @icewind1991 can confirm.

lib/FilesHooks.php
ArtificialOwl: this is improved in last commit

lib/FilesHooks.php
ArtificialOwl: I switched to `rootFolder->get(fullpath)`, but this might also limit the visibility on the access of the file as it limit the check on one path.

The fix was to merge result with the one from `$accessList['users']` that contains some other path to the targeted file.

please check last commit.

lib/FilesHooks.php
```diff
@@ -1219,4 +1212,160 @@ protected function commitActivityTransaction(bool $shouldFlush): void {
...
+					return $user->getUID();
+				}, $group->getUsers());
+
+				// merge current user list with members of the group
+				$usersToCheck = array_values(array_unique(array_merge($usersToCheck, $userIds)));
+			}
+		}
+
+
+		// now that we have a list of 'unstable' users, we confirm they have no access to the file
```
icewind1991, line 1354: "unstable" doesn't seem to be the correct word here?

icewind1991 (changes requested):
I'm not a fan of having additional explicit cross-app dependencies.

We can maybe add an storage extention interface to core that has a `filterUserAccess` or something similar to decouple things, but that can be done later.

Overall the approach of the filtering is fine and shouldn't be to bad for performance, once the listed issues are addressed this is good from my pov

lib/FilesHooks.php
```diff
@@ -1219,4 +1212,160 @@ protected function commitActivityTransaction(bool $shouldFlush): void {
...
+			if (!array_key_exists($cachedMount['visiblePath'], $knownRules[$storageId])) {
+				// we need mountPoint and folderId to generate the correct path
+				try {
+					$node = $this->rootFolder->get($fullPath);
+					$mountPoint = $node->getMountPoint();
+					$folderId = $mountPoint->getFolderId();
+					$path = substr($fullPath, strlen($mountPoint->getMountPoint()));
+				} catch (\Exception $e) {
+					// in case of issue during the process, we can imagine the user have no access to the file
+					$usersToCheck[] = $fullPath;
```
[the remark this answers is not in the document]
ArtificialOwl: wow.. thanks.

This is from before the implementation of `$cachedPath`.

lib/FilesHooks.php
```diff
@@ -1219,4 +1212,160 @@ protected function commitActivityTransaction(bool $shouldFlush): void {
...
+			}
+
+			$cachedPath[$cachedMount['userId']] = $fullPath = $cachedMount['path'];
+
+			// caching rules based on storage+path to file
+			if (!array_key_exists($cachedMount['visiblePath'], $knownRules[$storageId])) {
+				// we need mountPoint and folderId to generate the correct path
+				try {
+					$node = $this->rootFolder->get($fullPath);
+					$mountPoint = $node->getMountPoint();
```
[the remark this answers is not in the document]
ArtificialOwl: do you have any pointer for the easiest way to check ACL is enabled on mountpoint ?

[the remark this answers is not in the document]
ArtificialOwl: please check last commit regarding `GroupMountPoint` and `getFolderAclEnabled`

icewind1991 (approved):

lib/FilesHooks.php
PVince81: @nickvergessen are you ok with the last changes ? (asking since you initiated the topic)

nickvergessen: I have no clue on the actual content. Just was scared of getUserFolder() inside a loop. Seems to be gone now, so not an issue anymore.

lib/FilesHooks.php
```diff
@@ -1219,4 +1212,160 @@ protected function commitActivityTransaction(bool $shouldFlush): void {
...
+				}
+
+				$knownRules[$storageId][$cachedMount['visiblePath']] = true;
+			}
+		}
+
+		// using each rules that disable read permission to generate a list of users
+		// that might not have access to fileId
+		foreach ($rules as $rule) {
+			if (($rule->getMask() & 1) === 0
```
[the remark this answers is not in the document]
nickvergessen: This is still open?!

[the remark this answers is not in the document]
PVince81: @ArtificialOwl can you fix it?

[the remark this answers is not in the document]
ArtificialOwl: this was already fixed; conversation is resolved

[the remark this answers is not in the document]
ArtificialOwl: (my bad, there was 2 locations. pushing it)

ArtificialOwl:
- fixed the 2nd location for `Constants::PERMISSION_READ`
- squashed
- rebased

PVince81 (approved):
:+1:

fschrempf:
> We can maybe add an storage extention interface to core that has a filterUserAccess or something similar to decouple things, but that can be done later.

> once merged and backported we can look into make it cleaner if low effort

This is a real problem and it's one example for why NC code probably gets harder and harder to maintain. You are adding technical debt and "can be done later" and "make it cleaner if low effort" effectively means that it will never be done unless there's really someone who tracks these issues.

PR #1022, opened 2022-12-01
Revert "[stable23] Do not pass an instance of User where a string is expected"

blizzz:
Reverts nextcloud/activity#1021

Merged too early – we're in RC2.

PR #1023, opened 2022-12-01
Revert "Revert "[stable23] Do not pass an instance of User where a string is expected""

blizzz:
Reverts nextcloud/activity#1022

It was merged too early – reopening.


come-nc (approved):
Ok but please don’t revert the revert of the revert.

PR #1041, opened 2022-12-26
Adding the Auto-Submitted header for notification mails

bennet0496:
In PR nextcloud/server#35876, I added the possibility to directly set the `Auto-Submitted` mail header from `OCP\Mail\Message`. Reasons for the header are described in this PR and in an older issue nextcloud/server#16538 and PR  nextcloud/server#27200. But the short of it is that it prevents OOO/vacaction auto-reponses on properly configured mail system, that other wise would go to the configured sender address from the NC instance.

This PR adds utilization of the method introduced in nextcloud/server#35876, however there also is a fallback way via `getSwiftMessage()` allowing backwards compatibility for older NC version.


lib/MailQueueHandler.php
```diff
@@ -35,7 +35,7 @@
 use OCP\IURLGenerator;
 use OCP\IUser;
 use OCP\IUserManager;
-use OCP\L10N\IFactory;
+use OCP\L10N\IFactory;
```
nickvergessen, line 38:
```suggestion
use OCP\L10N\IFactory;
```

lib/MailQueueHandler.php
```diff
@@ -411,6 +411,12 @@ protected function sendEmailToUser($userName, $email, $lang, $timezone, $maxTime
 		$message->useTemplate($template);
 		$message->setFrom([$this->getSenderData('email') => $this->getSenderData('name')]);
 
+		if(method_exists($message, 'setAutoSubmitted')){
```
nickvergessen, line 414: Activity is bundled/shipped with the server so once the server PR is merged, this is not needed.

nickvergessen: But we can copy the else branch to stable25 and previous as a backport

bennet0496: Ok. And how would that work? creating new PRs against these older branches? or something else?

nickvergessen: That only makes sense if the server bit would be backported and I'm not sure it is, as it does not look like a critical nor security relevant thing

nickvergessen (approved):

susnux (approved):

Issue #1020, opened 2022-12-01
[BITV] List of activities should be marked up as `<ul>` unordered list with `<li>` elements


JuliaKirschenheuter:
Re-create a content of activities to semantic right `list`.

PR #1071, opened 2023-01-31
Create an unordered list for the list of activities

JuliaKirschenheuter:
Fixes https://github.com/nextcloud/activity/issues/1020

|  Before   | After                                                        |
|-------------|---------------------------------------------------------------|
| Screenshot from 2023-01-31 11-18-14 | Screenshot from 2023-01-31 14-41-45

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,6 +1,11 @@
 #container h2 {
 	margin-left: 24px;
-	margin-bottom: 16px;
+}
+
+.section-header {
+	display: block;
+	padding: 30px 30px 0 30px;
+	margin-bottom: 0;
 }
 
 .activitysubject {
--- a/js/script.js
+++ b/js/script.js
@@ -202,12 +202,13 @@ $(function(){
 					}
 				}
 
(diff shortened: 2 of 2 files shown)


nickvergessen:
~~That's only half of it.~~

Files sidebar already uses ul-li:
https://github.com/nextcloud/activity/blob/master/src/views/ActivityTab.vue#L32-L37

JuliaKirschenheuter:
Hi @michaelnissenbaum,

From the last review you mentioned that the content of activities should be a list and not just mixed content. I'm curious if i achieved it. In current case i have a complex content inside, not only some text (please see the screenshots above). Thank you!

JuliaKirschenheuter:
> ~That's only half of it.~

Ok, does it mean i should replace it on other places (please write me which ones)? *besides sidebar

nickvergessen:
> Ok, does it mean i should replace it on other places (please write me which ones)? *besides sidebar

Sidebar is already fixed, and those are the 2 places

Pytal (approved):

michaelnissenbaum:
Hi @JuliaKirschenheuter the new code looks good to me.

PR #1079, opened 2023-02-08
perf: Lazy load dashboard components

juliushaertl:
Save loading and evaluating larger amounts of javascript on the early page load.

Bundle contents for comparison

## Before
Screenshot 2023-02-08 at 17 52 12

## After
Screenshot 2023-02-08 at 17 50 41


nickvergessen (approved):
Seems to work

marcelklehr:
Can we do the same for https://github.com/nextcloud/activity/blob/master/src/sidebar.js ?

juliushaertl:
Yes, that would also be good :)

PR #1088, opened 2023-02-21
Update documentation about setRichSubject

come-nc:
It now computes a parsed version automatically.


come-nc:
ref https://github.com/nextcloud/server/pull/34807

nickvergessen (approved):

PR #1112, opened 2023-03-09
fix: Properly import the ActivityTab dynamically

juliusknorr:
dynamic imports return the imported component as { default }. This fixes loading the activity sidebar in files.


juliushaertl:
FYI @AndyScherzinger Might be good to still have the backport of this in rc2 as without the files activity sidebar is not working.

marcelklehr (approved):
huh, makes sense. I wonder why my tests didn't catch this :/

PR #1133, opened 2023-03-23
using getUserFolder

ArtificialOwl:


nickvergessen (changes requested):
Can you please add a useful commit messages and explain why this is necessary? What was not working before and it's working now? Is it fixing any bug? It's it improving or decreasing performance?

ArtificialOwl:
> Can you please add a useful commit messages and explain why this is necessary? What was not working before and it's working now? Is it fixing any bug? It's it improving or decreasing performance?

Yes, sorry about that, but I needed more testing before having it to be reviewed.
I edited the original message

nickvergessen:
So now you explained the what 👍🏼 
But I still don't know why 😕 
Any log message or anything that you are trying to solve?

What is needed to produce the problem? External store, groupfolders, ...?

ArtificialOwl:
@nickvergessen 

In NC23, the use of `rootFolder->get()` with an absolute path (`/userid/files/my_folder/my_file.txt`) does not work and need to be split in `userid`+`relativepath` to use it this way: `rootFolder->getUserFolder(userId)->get('/my_folder/my_file.txt')`

This patch extract userId and relative path from the absolute path and use the right syntax for `rootFolder`

nickvergessen (approved):
🐘

PR #1163, opened 2023-05-10
groupfolder activities on move/rename

ArtificialOwl:
retrieve the list of affectedUsers related to groupfolders, including ACL checks.

On file moving, it does the handling the same way a file is moved from a shared folder to another. 
Meaning that if a user gain or loose read rights during the moving process the file is said as _new_ or _deleted_

This is from the point of view of a user with no ACL limitation:
Selection_065

And this is from the point of view of a user that have not access to the folder a file is moved to/from:
Selection_066

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -353,7 +353,16 @@ class FilesHooks {
 			$this->moveCase = false;
 			return;
 		}
-		$this->oldAccessList = $this->getUserPathsFromPath($this->oldParentPath, $this->oldParentOwner);
+
+		$oldAccessList = $this->getUserPathsFromPath($this->oldParentPath, $this->oldParentOwner);
+
+		// file can be shared using GroupFolders, including ACL check
+		if ($this->config->getSystemValueBool('activity_use_cached_mountpoints', false)) {
+			[, , $oldFileId] = $this->getSourcePathAndOwner($oldPath);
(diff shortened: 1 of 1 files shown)


lib/FilesHooks.php
```diff
@@ -463,6 +478,12 @@ protected function fileMoving($oldPath, $newPath) {
 		$affectedUsers = $accessList['users'];
 		$oldUsers = $this->oldAccessList['users'];
 
+		// file can be shared using GroupFolders, including ACL check
+		if ($this->config->getSystemValueBool('activity_use_cached_mountpoints', false)) {
+			$this->userMountCache->clear(); // clear cache for new data
+			$affectedUsers = array_merge($affectedUsers, $this->getAffectedUsersFromCachedMounts($fileId));
```
artonge, line 484: Why merge? Isn't `getAffectedUsersFromCachedMounts` supposed to return an updated list?

artonge: Same question for other changes.

ArtificialOwl: the list is cached by UserMountCache. because we are moving files, we only get old data from pre-move; clearing the cache allows to get fresh data post-move

icewind1991 (approved):

artonge (approved):

PR #1165, opened 2023-05-11
Adapt test to modern getMount()

artonge:
`MountManager->getMount()` is now throwing an error instead of returning `null`.

This commit makes sure that we handle properly the cases of un-existing mount.

- https://github.com/nextcloud/server/pull/31431/files#diff-1a9e0726ff3c8a791a887953a207ebb95f3c51bb9fa3fdba365bed61747166c2R118
- https://github.com/nextcloud/server/pull/36836/files#diff-64a02ee1149d33d0747360694c2abf4ef1bdbfc379b4bb40f76c22b5b1fc06d9R203

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -34,7 +34,6 @@ use OCP\Activity\IManager;
 use OCP\Constants;
 use OCP\Files\Config\IUserMountCache;
 use OCP\Files\IRootFolder;
-use OCP\Files\Mount\IMountPoint;
 use OCP\Files\Node;
 use OCP\Files\NotFoundException;
 use OCP\IConfig;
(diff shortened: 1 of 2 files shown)


tests/bootstrap.php
```diff
@@ -24,7 +24,7 @@
 	define('PHPUNIT_RUN', 1);
 }
 
-require_once __DIR__.'/../../../lib/base.php';
+require_once __DIR__.'/../../../instances/server1/lib/base.php';
```
nickvergessen, line 27: ???

artonge:
Posted in the PR for Robin.

nickvergessen (approved):

PR #1166, opened 2023-05-12
Use the color-primary-element* variables

szaimen:
Explanation: the color-primary variables are not to be used in components because the introduce problems with high-contrast primary colors. Fix this by using the primary-element variables instead.


artonge (approved):

DUK3135:
Hey I was wondering if someone could help me out with some info about What
my options about a crypto currency company and how the server glitches and
one of my deposits comes ìn as a 8 million dollar transfer the company
seized my money put my account on suspension I tried to tell them then
won't let m withdraw and I put over 1500 and now these assholes got me
Locked out stating there fixing the issue and bugs associated with it

Issue #1110, opened 2023-03-08
[BITV] 9.4.1.2/5.4 - The "Activity settings" button is an accordion button that should be implemented as such with the aria-expanded attribute.


AndyScherzinger:
See https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/ for further details. (1)

- [ ] Fix
- [ ] Backport

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-bfbf07e5c3-v5-n4

</details>

PR #1230, opened 2023-06-28
Add more aria attributes to activity-settings button

Fenn-CS:
- Adds `aria-expanded` as this buttons triggers the expand or collapse of another section
- Add `aria-label` for better support of assitive technologies.

NB: The mechanism to change the state of the `aria-expanded` attribute, leaves out of this repository in https://github.com/nextcloud/server/blob/master/core/src/OC/apps.js

Resolves : https://github.com/nextcloud/activity/issues/1110

### DIFF
--- a/templates/stream.app.navigation.php
+++ b/templates/stream.app.navigation.php
@@ -41,7 +41,9 @@
 
 	<div id="app-settings">
 		<div id="app-settings-header">
-			<button class="settings-button" data-apps-slide-toggle="#app-settings-content"><?php p($l->t('Activity settings'));?></button>
+			<button class="settings-button" aria-expanded="false" aria-controls="app-settings-content" aria-label="Activity settings" data-apps-slide-toggle="#app-settings-content">
+				<?php p($l->t('Activity settings'));?>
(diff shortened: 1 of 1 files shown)


templates/stream.app.navigation.php
```diff
@@ -41,7 +41,9 @@
 
 	<div id="app-settings">
 		<div id="app-settings-header">
-			<button class="settings-button" data-apps-slide-toggle="#app-settings-content"><?php p($l->t('Activity settings'));?></button>
+			<button class="settings-button" aria-expanded="false" aria-label="Activity settings" data-apps-slide-toggle="#app-settings-content">
```
Fenn-CS, line 44: The state change is handled elsewhere. 

See : https://github.com/nextcloud/server/pull/39049

susnux:
If there is something expanded shouldn't there be a aria-controls property?

artonge (approved):

Fenn-CS:
Right @susnux, fixed. Thank you.

susnux (approved):

Issue #1056, opened 2023-01-20
Trying to access array offset on value of type null at /lib/FilesHooks.php#471


solracsf:
Nextcloud 24.0.9.

```
{
  "reqId": "Ln4zf0y4OwEBrrxQBull",
  "level": 3,
  "time": "2023-01-20T07:59:14+00:00",
  "remoteAddr": "77.201.80.209",
  "user": "larissa.assi@cabinet-lacomblez.fr",
  "app": "PHP",
  "method": "MOVE",
  "url": "/remote.php/dav/files/laris/FOLDER%201/FOLDER%202",
  "message": "Trying to access array offset on value of type null at /var/www/nextcloud/apps/activity/lib/FilesHooks.php#471",
  "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36",
  "version": "24.0.9.2",
  "exception": {
    "Exception": "Error",
    "Message": "Trying to access array offset on value of type null at /var/www/nextcloud/apps/activity/lib/FilesHooks.php#471",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/nextcloud/apps/activity/lib/FilesHooks.php",
        "line": 471,
        "function": "onError",
        "class": "OC\\Log\\ErrorHandler",
        "type": "::",
        "args": [
          2,
          "Trying to access array offset on value of type null",
          "/var/www/nextcloud/apps/activity/lib/FilesHooks.php",
          471
        ]
      },
      {
        "file": "/var/www/nextcloud/apps/activity/lib/FilesHooks.php",
        "line": 386,
        "function": "fileMoving",
        "class": "OCA\\Activity\\FilesHooks",
        "type": "->",
        "args": [
          "/FOLDER 1/FOLDER 2",
          "/FOLDER 1/FOLDER 3/FOLDER 4"
        ]
      },
      {
        "file": "/var/www/nextcloud/apps/activity/lib/FilesHooksStatic.php",
        "line": 79,
        "function": "fileMovePost",
        "class": "OCA\\Activity\\FilesHooks",
        "type": "->",
        "args": [
          "/FOLDER 1/FOLDER 2",
          "/FOLDER 1/FOLDER 3/FOLDER 4"
        ]
      },
      {
        "file": "/var/www/nextcloud/lib/private/legacy/OC_Hook.php",
        "line": 106,
[...]

PR #1281, opened 2023-08-12
Check if users is set

solracsf:
Error: Undefined array key "users"
Exception: array_keys(): Argument #1 ($array) must be of type array, null given

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -151,7 +151,7 @@ class FilesHooks {
 			$this->generateRemoteActivity($accessList['remotes'], $activityType, time(), $this->currentUser->getCloudId(), $accessList['ownerPath']);
 		}
 
-		$affectedUsers = $accessList['users'];
+		$affectedUsers = $accessList['users'] ?? [];
 
 		// file can be shared using GroupFolders, including ACL check
 		if ($this->config->getSystemValueBool('activity_use_cached_mountpoints', false)) {


artonge (approved):

artonge:
In which case can it be empty? Do you have reproduction steps?

solracsf:
No sorry, just some errors like these in the logs...

artonge:
I am wondering if this won't juste hide another issue. Can you investigate ?

artonge:
Any news @solracsf ?

solracsf:
I'm wondering if 

`$affectedUsers = $accessList['users'] ?? [];`

could be a better solution here. Passing an empty array in case of `null` would not invalidate further logic in the `array_merge` and fix the "not an array" exception. 🤔

artonge:
Maybe. Do you still have the log?

lib/FilesHooks.php
```diff
@@ -231,34 +231,36 @@ protected function addNotificationsForFileAction($filePath, $activityType, $subj
...
-		foreach ($affectedUsers as $user => $path) {
-			$user = (string)$user;
-
-			if ($user === $this->currentUser->getUID()) {
-				$userSubject = $subject;
-				$userParams = [[$fileId => $path]];
-			} else {
-				$userSubject = $subjectBy;
-				$userParams = [[$fileId => $path], $this->currentUser->getUserIdentifier()];
+		if (!empty($accessList['users'])) {
```
nickvergessen, line 234: This is not correct. You are skipping the `$this->config->getSystemValueBool('activity_use_cached_mountpoints', false)` block ❗ 

The actual fix would be to:
```php
$affectedUsers = $accessList['users'] ?? [];
```
on line 234

nickvergessen (changes requested):

lib/FilesHooks.php
solracsf: Godd spot. Fixed, thanks.

solracsf:
This will also fix:

```
array_keys(): Argument #1 ($array) must be of type array, null given in file '/var/www/nextcloud/apps/activity/lib/FilesHooks.php' line 243"
```
Can we go ahead and merge + backport it?

susnux:
> I think this should be fixed in `getUserPathsFromPath` because there are other places where it is used

I think this is still valid  to not just hide issues

Messj1:
@ solracsf  Is this problem is related to #1056 ? Can you Check if following commit helps: 8facb65a39e4a49e11255acd417e06f3ec016f45

solracsf:
Maybe, but as long as I don't see this in logs a long time ago, and this PR is not the good approach anyway, closing here 👍

Issue #1280, opened 2023-08-11
Could not boot activity: Call to undefined method OC\\Server::getEventDispatcher()


skjnldsv:
```json
{
  "reqId": "Q6Pcso2vbb8oj6hmhkCL",
  "level": 4,
  "time": "2023-08-11T18:17:34+00:00",
  "remoteAddr": "172.18.0.1",
  "user": "admin",
  "app": "no app in context",
  "method": "GET",
  "url": "/ocs/v2.php/apps/notifications/api/v2/notifications",
  "message": "Could not boot activity: Call to undefined method OC\\Server::getEventDispatcher()",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36",
  "version": "28.0.0.2",
  "exception": {
    "Exception": "Error",
    "Message": "Call to undefined method OC\\Server::getEventDispatcher()",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/nextcloud/apps2/activity/lib/AppInfo/Application.php",
        "line": 143,
        "function": "registerFilesActivity",
        "class": "OCA\\Activity\\AppInfo\\Application",
        "type": "->",
        "args": []
      },
      {
        "file": "/var/www/nextcloud/lib/private/AppFramework/Bootstrap/Coordinator.php",
        "line": 200,
        "function": "boot",
        "class": "OCA\\Activity\\AppInfo\\Application",
        "type": "->",
        "args": [
          [
            "OC\\AppFramework\\Bootstrap\\BootContext"
          ]
        ]
      },
      {
        "file": "/var/www/nextcloud/lib/private/App/AppManager.php",
        "line": 433,
        "function": "bootApp",
        "class": "OC\\AppFramework\\Bootstrap\\Coordinator",
        "type": "->",
        "args": [
          "activity"
        ]
      },
      {
        "file": "/var/www/nextcloud/lib/private/App/AppManager.php",
        "line": 212,
        "function": "loadApp",
        "class": "OC\\App\\AppManager",
        "type": "->",
        "args": [
          "activity"
        ]
      },
      {
        "file": "/var/www/nextcloud/lib/private/legacy/OC_App.php",
        "line": 126,
        "function": "loadApps",
        "class": "OC\\App\\AppManager",
        "type": "->",
        "args": [
          []
        ]
      },
      {
[...]

PR #1285, opened 2023-08-18
Drop Symfony event dispatcher

artonge:
Fix https://github.com/nextcloud/activity/issues/1280


lib/Event/LoadAdditionalScriptsEvent.php
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * @since 28.0.0 Dispatched as a typed event
+ */
+class LoadAdditionalScriptsEvent extends Event {
+	/**
+	 * @deprecated 28.0.0 - Listen to the typed event instead.
+	 */
+	public const EVENT_ENTITY = 'OCA\Activity::loadAdditionalScripts';
+
+	public function __construct(public readonly string $filter) {
```
nickvergessen, line 35: readonly is 8.1+ and it's not yet 100% clear 8.0 will be dropped.

nickvergessen (approved):
Good apart from the readonly

artonge:
Ah, forgot to pull before pushing :see_no_evil:

lib/Event/LoadAdditionalScriptsEvent.php
nickvergessen: This is back because you dropped https://github.com/nextcloud/activity/commit/dd4805cf9206e2cb6a7ea7b8a1158ae33ae3b611

lib/Listener/ShareEventListener.php
```diff
@@ -0,0 +1,66 @@
...
+namespace OCA\Activity\Listener;
+
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\Share\Events\BeforeShareDeletedEvent;
+use OCP\Share\Events\ShareDeletedFromSelfEvent;
+
+/**
+ * The class to handle the share events
+ * @template-implements IEventListener<\OCP\Share\Events\BeforeShareDeletedEvent|\OCP\Share\Events\ShareDeletedFromSelfEvent>
```
nickvergessen, line 32: It's a bit weird, this has to be:

```suggestion
 * @template-implements IEventListener<\OCP\EventDispatcher\Event>
```

nickvergessen:
So, force merge because cypress is fixed in the other PR?

artonge:
Or rebase :)

PR #1295, opened 2023-08-28
Fix Cypress tests for F2V

artonge:
After the F2V rewrite, some tests needed to be adapted.


nickvergessen (approved):
CI is green, so should be fine

PR #1301, opened 2023-09-07
[stable27] chore(deps): Downgrade nextcloud/ocp to stable27

ChristophWurst:
This fix is needed for https://github.com/nextcloud/activity/pull/1299#issuecomment-1708377412.

- [x] Requires https://github.com/nextcloud/activity/pull/1304

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
 		]
 	},
 	"require-dev": {
-		"nextcloud/ocp": "dev-master",
+		"nextcloud/ocp": "dev-stable27",
 		"phpunit/phpunit": "^9.5",
 		"psalm/phar": "^5.9"
 	},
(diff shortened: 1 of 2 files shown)


ChristophWurst:
> Class, interface or enum named Symfony\Component\EventDispatcher\EventDispatcherInterface

artonge:
I fixed this somewhere else by using the phar package of psalm. Not sure if good practice, but it works.

ChristophWurst:
> friendsofphp/php-cs-fixer v3.13.2 requires  symfony/event-dispatcher (^5.4 || ^6.0) 

hence genericevent etc are found on stable27 but not here. I'll add those extraneous classes to the psalm ignore list

miaulalala (approved):

PR #1303, opened 2023-09-08
chore(deps): Move cs-fixer into a composer bin

ChristophWurst:
Contributes to https://github.com/nextcloud/activity/issues/1300

PR #1304, opened 2023-09-08
[stable27] Resolve dependency hell

ChristophWurst:
nextcloud/ocp is outdated. It has the master ref but can't be updated to latest. It can also not be changed to stable27. It's because of a psr/container conflict between nextcloud/ocp, cs-fixer and psalm.

- Backport https://github.com/nextcloud/activity/pull/1303
- Use phar version of psalm

With these changes applied it's possible to change nextcloud/ocp to dev-stable27


miaulalala (approved):
Thanks for this!

PR #1309, opened 2023-09-11
fix(db): pre- SELECT ids with a LIMIT to control the amount of rows deleted for MySQL

miaulalala:
Doctrine does not allow LIMITs on DELETE queries:

https://stackoverflow.com/questions/25308171/doctrine-query-delete-with-limit

Thus we need to handle MySQL/MariaDB separately as clustered setups are not going to be able to handle more than 128k rows / 2GB of writes at any one time.

I have set the current LIMIT to 10k but an argument could probably be made to use 20k. We can adjust in the future. I choose this limit to handle clustered setups where wsrep_max_ws_size or wsrep_max_ws_rows is lower than the expected value.

MySQL also doesn't allow LIMITs in IN subqueries.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -59,6 +59,7 @@ use OCP\User\Events\PostLoginEvent;
 use OCP\User\Events\UserDeletedEvent;
 use OCP\Util;
 use Psr\Container\ContainerInterface;
+use Psr\Log\LoggerInterface;
 
 class Application extends App implements IBootstrap {
 	public const APP_ID = 'activity';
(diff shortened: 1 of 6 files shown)


miaulalala:
ok this is weird - MariaDB inherits from MySQL anyway. Lemme investigate.

kesselb (approved):

nickvergessen:
> Doctrine does not allow LIMITs on DELETE queries:

Can we add a error (log) in the querybuilder when one tries that. I guess it's used in more places

nickvergessen:
> it will throw anyway from doctrine

But the activity app never logged an error from it's cron job on our instance?

PR #1341, opened 2023-09-28
Remove click functionality on row entry

JuliaKirschenheuter:
Resolves https://github.com/nextcloud/server/issues/40672

Before:

Peek 2023-09-28 09-32

After:

Peek 2023-09-28 10-28


src/components/ActivityGrid.vue
```diff
@@ -38,7 +38,7 @@
 			<tr v-for="(activity, activityKey) in group.activities" :key="activityKey">
 				<th scope="row">
 					<!-- eslint-disable vue/no-v-html -->
-					<label @click="toggleMethodsForActivity({groupKey, activityKey})" v-html="activity.desc" />
+					<label v-html="activity.desc" />
```
ShGKme, line 41: This method was used only here, so we can also remove it from the component's methods (`mapActions`, in line 85).

But keeping the method in the store, if case we add a "check all" button later.

ShGKme: Should we make this element just a `span`/`div`, as this `<label>` has no linked elements?

artonge: Or shall we link it to the checkbox ?

ShGKme: > Or shall we link it to the checkbox ?

We cannot, there are 2 checkboxes

ShGKme (approved):

Pytal (approved):

marcoambrosini (approved):

susnux (approved):

PR #1361, opened 2023-10-11
fix(cypress): Make sure cypress tests work with upcoming Nextcloud 28

susnux:
(or: Fix cypress on master branch)


Pytal (approved):

artonge (approved):

susnux:
Seems to fix all but activities, because after the click on the menu toggle the files actions menu is not opened fast enough. Let me try something.

BTW this is CI only as you machine (or my) is faster than the CI runners...

susnux:
Well yes fixed by waiting for the popover to be visible :heavy_check_mark: :partying_face:

PR #1363, opened 2023-10-12
fix(sidebar): Only load chunk once

susnux:
Only load the chunk if needed (yes it might be cached but we do not need to try this).


artonge (approved):
So it was loaded whenever the sidebar was opened ?

susnux:
> So it was loaded whenever the sidebar was opened ?

1. when the DOM was loaded
2. when the Sidemenu as registered, so the second one is not needed

src/sidebar.js
```diff
@@ -41,17 +41,21 @@ const activityTab = new OCA.Files.Sidebar.Tab({
 	iconSvg: LightningBolt,
 
 	async mount(el, fileInfo, context) {
-		const { default: ActivityTab } = await import(/* webpackPreload: true */ './views/ActivityTab.vue')
```
marcelklehr, line 44: Huh, webpack doesn't cache the chunk? That's weird.

PR #1394, opened 2023-11-05
fix: Add stub for files event to fix psalm error

susnux:
Fixing:

```
ERROR: InvalidArgument - lib/AppInfo/Application.php:139:13 - Incompatible types found for T (OCA\Files\Event\LoadSidebar|OCP\EventDispatcher\Event is not in OCP\EventDispatcher\Event) (see https://psalm.dev/004)
                $context->registerEventListener(LoadSidebar::class, LoadSidebarScripts::class);
```

### DIFF
--- a/tests/stubs/oca_files_event.php
+++ b/tests/stubs/oca_files_event.php
@@ -1,0 +1,6 @@
+<?php
+
+namespace OCA\Files\Event {
+	class LoadSidebar extends \OCP\EventDispatcher\Event {
+	}
+}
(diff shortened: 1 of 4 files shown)


tests/stubs/oca_files_event.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+namespace OCA\Files\Event {
+	class LoadSidebar extends \OCP\EventDispatcher\Event {
+	}
+}
```
nickvergessen, line 6:
```suggestion
}

```

artonge (approved):

susnux:
Cypress unrelated, seems to be broken on master branch.
So guess it can be merged by someone with permissions :)

artonge:
No

susnux:
In this case I try to fix Cypress as well

PR #1409, opened 2023-11-12
Cleanup repository

susnux:
* Remove unneeded handlebars scrips
* Fix stylelint config and add CI workflow for it
* Move JS mocks so they are located next to the tests
* Migrate deprecated `::v-deep` to `:deep`


src/__tests__/Activity.test.ts
```diff
@@ -24,9 +24,9 @@ import { afterAll, beforeAll, expect, test, vi } from 'vitest'
 import { nextTick } from 'vue'
 
 import Activity from '../components/Activity.vue'
-import ActivityModel from '../models/ActivityModel.ts'
```
artonge, line 27: IIRC, we agreed to keep file extensions in imports, so please add it back :)

artonge (changes requested):

src/__tests__/Activity.test.ts
susnux: Yes, fixed it.

artonge (approved):

susnux:
> Node CI is failing a lot recently. Has there been a change in node or npm version?

No there is something weird going on, the assets I compile locally slightly differ from what the CI is compiling.
But I use exactly the same NPM and node version, same dependencies (npm ci).

PR #1396, opened 2023-11-08
POC: Minimal example of actions API

susnux:
This is a minimal example how an API could look like to register actions for the Activity sidebar.
For example it could be used to register the comment input action on the activity tab.

See code or this example how the usage would look like:
```js
/**
 * Dummy example of how to add a sidebar action
 */
window.OCA.Activity.registerSidebarAction({
	mount: (el) => {
		// eslint-disable-next-line no-console
		console.log('Register example plugin')
		const label = document.createElement('label')
		label.innerText = 'Write a comment'
		label.appendChild(document.createElement('input'))
		label.appendChild(document.createElement('hr'))
		label.style.display = 'flex'
		label.style.flexDirection = 'column'
		el.appendChild(label)
	},
	// eslint-disable-next-line no-console
	unmount: () => console.log('unregister example plugin'),
})

window.OCA.Activity.registerAction('comments', {
	label: 'Delete comment',
	handler: (activity) => console.log(activity),
})
```

This example looks bad, because no styling is done but it works:
image

Of cause you could simply mount the comment app's input component, no big adjustments needed.

PR #1416, opened 2023-11-15
feat(sidebar): Add frontend API that allows to register custom entries

### LINKED ISSUE #278: Improvements for Comments in Activity
As commented in https://github.com/nextcloud/android/pull/2690, there's some improvements we can do to how Comments show in the activity stream:

- The first line of a comment »Commented by you« is not really useful. It should simply say »You« instead (or the person's name)
- Similar for the icon: Instead of the generic comment icon, it should show the persons avatar

cc @nickvergessen

susnux:
* Resolves #278 
* Replaces #1396 

This adds a front end API that allows to register plugins on the Activity tab:
1. Register custom actions, this actions will be shown on top of the activity bar (e.g. a comments input)
2. Register custom stream elements, meaning custom entries in the activity stream (e.g. comments)
3. Register custom filter function for shown activities to filter out some

Why showing custom elements instead of the activity? Activities will be cleaned up while e.g. comments do not. So this allows e.g. comments to be shown within the activity sidebar tab.


susnux:
The diff looks awful but in reality e.g. Activity.vue was just renamed to GenericActivity.vue and Activity.vue is now a wrapper...

susnux:
A screen cast of how it looks when comments use it can be found here:
https://github.com/nextcloud/server/pull/41491

AndyScherzinger (approved):

artonge (approved):

PR #1438, opened 2023-11-29
fix(sidebar): Refactor code to also load plugin entries when Activity API returns 304

susnux:
* Resolves https://github.com/nextcloud/server/issues/41791

We should load activities separate from other entries as 304 is no error but will throw an exception and thus skip loading extra entries (comments).


ChristophWurst (approved):

Pytal (approved):

artonge (approved):

susnux:
/backport ca31c7812b07f63cb5e6afdab612da6e6cce92f1 to stable28

PR #1468, opened 2023-12-18
Fix missing parameters when type is not "known"

nickvergessen:
Currently the fall backs from https://github.com/nextcloud/activity/blob/stable27/js/richObjectStringParser.js#L82-L87 are missing breaking activity of calendar, deck, talk, …

Before | After
---|---
Bildschirmfoto vom 2023-12-18 15-44-37 | Bildschirmfoto vom 2023-12-18 15-44-45


nickvergessen:
/backport 2a2939fd58dec7ac24d0622dd9c5b7a4e676c135,f77b042f423308688dd3e5d51840d80d9ff7126b to stable28

src/views/ActivityAppNavigation.vue
```diff
@@ -49,14 +49,11 @@ import { translate as t } from '@nextcloud/l10n'
...
-import {
-	NcAppNavigation,
-	NcAppNavigationItem,
-	NcAppNavigationSettings,
-	NcButton,
-	NcCheckboxRadioSwitch,
-} from '@nextcloud/vue'
-
+import NcAppNavigation from '@nextcloud/vue/dist/Components/NcAppNavigation.js'
+import NcAppNavigationItem from '@nextcloud/vue/dist/Components/NcAppNavigationItem.js'
```
susnux, line 53: Any reason?

susnux (approved):

artonge (approved):

src/views/ActivityAppNavigation.vue
nickvergessen: On master the js/ size is 8.6M, with this small adjustment we only import what we need, bringing the package size down to 6.0M in js/

bigcat88 (approved):

artonge:
Compilation CI seem off on this repo, not sure why

susnux:
> Compilation CI seem off on this repo, not sure why

If you compile with the compile command it works, if you compile locally it does not :/

artonge:
> If you compile with the compile command it works, if you compile locally it does not :/

From my experience, the compile command also fails.
Do you know the reason?

nickvergessen:
The main difference to other repos is this here uses vite not eslint for building?
As per https://github.com/nextcloud/activity/actions/runs/7249713557/job/19748361102#step:8:8 that seems to rename artefacts 🙈

susnux:
> that seems to rename 

Assets are named by the content hash thus the content seems to be different between those builds

PR #1519, opened 2024-01-17
fix: Replace deprecated function calls and private with public API

susnux:
* Replace deprecated `execute` with `executeQuery` or `executeStatement`.
* Replace private `OC\BackgroundJob\...` with public `OCP\BackgroundJob\...`
* Do not use private `\OC::$server` but use dependency injection


artonge (approved):

miaulalala (approved):

nickvergessen (commented):
💥 This breaks the digest cron  job

PR #1536, opened 2024-01-23
fix rss feed item guid

mcnesium:
I am trying to use the RSS feed feature of this application. For example, it could be useful for [embedding it in a Matrix team collaboration room](https://matrix-org.github.io/matrix-hookshot/latest/setup/feeds.html) so that everyone in the room can immediately see when a team member has worked on an item. In less busy environments, it might also help to subscribe to the feed in your regular everyday feed reader and be notified of activity that way. Both help to avoid having to actively check the activity application in the Nextcloud browser interface.

I have been doing this for a couple of weeks now, but I have been very unhappy with the output so far. The RSS feed items just made no sense. After a file had been created or edited, file modification notices would appear at what seemed to me to be random times, they would reappear for days on end, even though no one had touched the file. I didn't file a bug about this because I just couldn't describe what was wrong. I could not see any patterns in the feed output and it was very difficult to figure out which RSS feed item was actually caused by which action on a file.

Then I got the idea to test the feed for validity, and voila, the [W3C Feed Validator](https://validator.w3.org/feed/) was unhappy with the item's `guid`:

So I checked a Wordpress feed and found that they validate fine. The relevant difference is that the contents of the `guid` element had a URL before an actual ID. Furthermore, [in the RSS specification](https://www.rssboard.org/rss-specification#ltguidgtSubelementOfLtitemgt) the `guid` has a URL part, too. So I edited `templates/rss.php` and literally hardcoded some URL into it.  I then monitored the output of my Matrix RSS feed bot and my regular feed reader. Suddenly, not only did they match, they even made sense :D

I can definitely not explain, *why* it is better with the URL part than just the `activity_id` in the `guid` element, but now it works for me.

This PR does not have a hardcoded URL in it but takes the `p($_['rssLink']);` function that is already in use in that template. The output now looks sort of like this:

          <guid isPermaLink="false">https://cloud.domain.de/apps/activity/rss.php?id=3720675</guid>

While that URL does not really lead anywhere, it does fix the weirdness of the RSS feed output in general. It could be improved by actually pointing to a visible event item in the browser frontend of this app, e.g. `…/apps/activity?id=3720675`. But that would require some `<a id="[activity_id]">…` HTML in the browser app template plus some function that returns that URL. However, I couldn't easily find out how to do this, also it would probably be beyond the scope of this PR.


artonge (approved):
So the guid must be globally unique and not only relative to the domain?
Nice find, thanks :)

artonge:
Can you fix the DCO ?

templates/rss.php
```diff
@@ -34,7 +34,7 @@
 		<atom:link href="<?php p($_['rssLink']); ?>" rel="self" type="application/rss+xml" />
 <?php foreach ($_['activities'] as $activity) { ?>
 		<item>
-			<guid isPermaLink="false"><?php p($activity['activity_id']); ?></guid>
+			<guid isPermaLink="false"><?php p($_['rssLink']); echo "?id="; p($activity['activity_id']); ?></guid>
```
susnux, line 37:
```suggestion
			<guid isPermaLink="false"><?php p($_['rssLink'] . '?id=' . $activity['activity_id']); ?></guid>
```

susnux (approved):

mcnesium:
Looks like the test needs to be updated as well :partying_face: 

Should I just add the `http://nextcloud.org?id=` to the string [here](https://github.com/nextcloud/activity/blob/master/tests/Template/RssTest.php#L81), [here](https://github.com/nextcloud/activity/blob/master/tests/Template/RssTest.php#L89) and [here](https://github.com/nextcloud/activity/blob/master/tests/Template/RssTest.php#L101)?

susnux:
> Should I just add

Yes please :)

mcnesium:
:tada:

PR #1545, opened 2024-02-02
MailQueueHandler: Catch provider errors

marcelklehr:


lib/MailQueueHandler.php
```diff
@@ -434,7 +434,7 @@ protected function parseEvent($lang, IEvent $event) {
 		foreach ($this->activityManager->getProviders() as $provider) {
 			try {
 				$event = $provider->parse($lang, $event);
-			} catch (\InvalidArgumentException $e) {
+			} catch (\Throwable $e) {
```
artonge, line 437: Should we log the exception?
Can you give some explanation for the change?

nickvergessen (changes requested):
At least without logging this is a very bad idea. It makes it impossible to find future errors by any app in their activity handling.

lib/MailQueueHandler.php
nickvergessen: https://github.com/nextcloud/deck/pull/5533 + https://github.com/nextcloud/deck/pull/5542

But yeah also not a fan of it and I think we should not do this.

marcelklehr: Added logging

marcelklehr: The reason for this change is that a single error in some app can block the sending of all activity emails.

artonge (approved):

nickvergessen (approved):

marcelklehr:
It seems like part of the activity API is to allow providers to throw \InvalidArgumentExceptions when they don't recognize an event, which are now flooding the logs, because they are part of \Throwable.

marcelklehr:
opened a PR: https://github.com/nextcloud/activity/pull/1586

PR #1586, opened 2024-03-06
MailQueueHandler: Don't flood logs with useless messages

marcelklehr:


lib/MailQueueHandler.php
```diff
@@ -435,6 +435,10 @@ protected function parseEvent($lang, IEvent $event) {
 			try {
 				$event = $provider->parse($lang, $event);
 			} catch (\Throwable $e) {
+				if ($e instanceof \InvalidArgumentException) {
+					// \InvalidArgumentException is part of the activity API, so we let it through
```
nickvergessen, line 439: This in itself is already an issue which I noticed in notifications recently. But it's something for later

nickvergessen (approved):

lib/MailQueueHandler.php
```diff
@@ -435,6 +435,10 @@ protected function parseEvent($lang, IEvent $event) {
...
+					throw $e;
+				}
```
come-nc, line 441:
```suggestion
			} catch (\InvalidArgumentException $e) {
				/* Ignore */
			} catch (\Throwable $e) {
```
Like this it would restore previous behavior of ignoring those errors silently.
If you throw them where are they catched?

marcelklehr: the method is documented to throw InvalidArgumentException, but I agree it's cleaner to not rethrow them.

come-nc (approved):

lib/MailQueueHandler.php
amenolo: Thanks for the patch, I just happen to notice that...

Before fix
image

After fix
image


NC 27.1.7 with following enabled apps

```
  - activity: 2.19.0
  - bookmarks: 13.1.3
  - cfg_share_links: 4.2.0
  - cloud_federation_api: 1.10.0
  - dav: 1.27.0
  - epubviewer: 1.6.0
  - federatedfilesharing: 1.17.0
  - files: 1.22.0
  - files_downloadlimit: 1.1.0
  - files_external: 1.19.0
  - files_pdfviewer: 2.8.0
  - files_rightclick: 1.6.0
  - files_sharing: 1.19.0
  - files_trashbin: 1.17.0
  - logreader: 2.12.0
  - lookup_server_connector: 1.15.0
  - nextcloud_announcements: 1.16.0
  - notifications: 2.15.0
  - oauth2: 1.15.2
  - occweb: 0.1.1
  - onlyoffice: 8.2.4
  - provisioning_api: 1.17.0
  - richdocuments: 8.2.5
  - serverinfo: 1.17.0
  - settings: 1.9.0
  - sharebymail: 1.17.0
  - snappymail: 2.35.3
  - text: 3.8.0
  - theming: 2.2.0
  - theming_customcss: 1.15.0
  - twofactor_backupcodes: 1.16.0
  - updatenotification: 1.17.0
  - viewer: 2.1.0
  - workflowengine: 2.9.0
```

It's just me or...? :confused:

PR #1635, opened 2024-04-11
Exclude specified users from activity expiration

akhil1508:
- On my nextcloud setup, I want to keep activities only for a few  months for all users
- However, the admin activities should be there forever as they provide a great way of keeping track of admin actions and also user actions that set the activity user as admin


lib/Data.php
```diff
@@ -48,7 +49,8 @@ class Data {
 	public function __construct(
 		protected IManager $activityManager,
 		protected IDBConnection $connection,
-		protected LoggerInterface $logger) {
+		protected LoggerInterface $logger,
+		protected IConfig $config) {
```
artonge, line 53:
```suggestion
		protected IConfig $config,
	) {
```

artonge (approved):
Looks good

artonge (commented):
Could you take a look at the CI errors?

akhil1508:
> Could you take a look at the CI errors?


@artonge Apologies, I didn't run PHP CS Fixer; should be good now

akhil1508:
@artonge  I don't get 

> Error: lib/AppInfo/Application.php:114:11: TooFewArguments: Too few arguments for OCA\Activity\Data::__construct - expecting config to be passed (see https://psalm.dev/025) 

As I didn't change `lib/AppInfo/Application.php`

miaulalala:
> @artonge I don't understand
> 
> > Error: lib/AppInfo/Application.php:114:11: TooFewArguments: Too few arguments for OCA\Activity\Data::__construct - expecting config to be passed (see [psalm.dev/025](https://psalm.dev/025))
> 
> As I didn't change `lib/AppInfo/Application.php`

You need to register the IConfig injection to 
https://github.com/nextcloud/activity/blob/59e65d4bce3d3e22548191185fb3a307452f58d5/lib/AppInfo/Application.php#L113-L119

```php
		$context->registerService(Data::class, function (ContainerInterface $c) {
			return new Data(
				$c->get(IManager::class),
				$c->get('ActivityConnectionAdapter'),
				$c->get(LoggerInterface::class),
				$c->get(IConfig::class),
			);
		});
```

artonge:
Could you also squash your commits into one?

akhil1508:
> Could you also squash your commits into one?

@artonge done

miaulalala:
Tests need the new constructor properties :)

lib/Data.php
```diff
@@ -347,11 +350,26 @@ public function validateFilter($filterValue) {
...
+		$conditions = [
 			'timestamp' => [$timelimit, '<'],
-		]);
+		];
+
+		$excludedUsers = $this->config->getSystemValue('activity_expire_exclude_users', []);
+		if (!empty($excludedUsers)) {
+			foreach ($excludedUsers as $user) {
+				$condition1 = [
+					'user' => [$user, '!=']
```
nickvergessen, line 362: should be only the `affecteduser` below.
This here would mean that actions of the given user would never expire.

Also this breaks the index usage of an already problematic query. So should only be used with care

akhil1508: @nickvergessen I'm confused, is the `affecteduser` the user for whom the activity is fired off or is the `affecteduser` the user the user who the activity is for?

If a user deleted an account and an activity for admins is fired, would the `affecteduser` row be the deleted username or the admin username? I want all the activities that are fired  for admin users to not expire, to maintain a history of important activities..

akhil1508:
> Tests need the new constructor properties :)

@miaulalala Done. I will squash again and fix commit push after final approval(s) to avoid doing it many times.

lib/Data.php
nickvergessen: User is the actor of the action,
Affected user is the one viewing the activity stream

But then it shows a bit that the config option needs better explanation.
But I still think that only 1 of the checks should stay

artonge:
Still one minor lint error

lib/Data.php
nickvergessen: > I want all the activities that are fired for admin users to not expire, to maintain a history of important activities..

Btw there is also the audit log feature in Nextcloud which is specifically designed for things like this:
https://docs.nextcloud.com/server/latest/developer_manual/basics/logging.html#admin-audit-logging

akhil1508: @nickvergessen But then I need to add some new code or patch the files on server as admin audit logs these user deletion and other important actions at "INFO" level.. It uses the `loglevel` variable and if I allow "INFO" level logs, I'm afraid performance takes a hit for my server :(

akhil1508: > should be only the affecteduser below.

@nickvergessen Done

akhil1508:
> Still one minor lint error

@artonge Fixed

lib/Data.php
nickvergessen: https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/logging_configuration.html#log-level-interaction

artonge:
> @artonge Fixed

Almost :)

miaulalala:
[akhil1508](https://github.com/akhil1508) do you still wanna work on this? Otherwise I'd take over :)

akhil1508:
> do you still wanna work on this? Otherwise I'd take over :)

@miaulalala Please take over, I might not be able to work on this for a while :|

miaulalala:
Thanks a lot for your work on this @akhil1508 !

PR #1700, opened 2024-06-06
Make container view looks proportional on viewport

solracsf:
On big screen resolution, actual implementation with fixed with looks like an iframe.
Make it fluid and align with other items on screen.

## Before
image

## After
image

### DIFF
--- a/src/views/ActivityAppFeed.vue
+++ b/src/views/ActivityAppFeed.vue
@@ -208,12 +208,10 @@ watch(props, () => {
 	&__container {
 		display: flex;
 		flex-direction: column;
-
 		height: 100%;
-		width: min(100%, 924px);
-		max-width: 924px;
-		margin: 0 auto;
-		padding-inline: 12px;
+		// Align with app navigation toggle
+		margin: var(--app-navigation-padding, 8px) 0 0 calc(2 * var(--app-navigation-padding, 8px) + 44px)
+		padding-inline: 0 44px;
 		overflow-y: scroll;
 	}
 


src/views/ActivityAppFeed.vue
```diff
@@ -208,12 +208,10 @@ watch(props, () => {
...
 		display: flex;
 		flex-direction: column;
-
 		height: 100%;
-		width: min(100%, 924px);
-		max-width: 924px;
-		margin: 0 auto;
-		padding-inline: 12px;
+		// Align with app navigation toggle
+		margin: var(--app-navigation-padding, 8px) 0 0 calc(2 * var(--app-navigation-padding, 8px) + 44px);
```
ShGKme, line 213:
```suggestion
		margin: var(--app-navigation-padding, 8px) 0 0 calc(2 * var(--app-navigation-padding, 8px) + var(--default-clickable-area));
```

src/views/ActivityAppFeed.vue
```diff
@@ -208,12 +208,10 @@ watch(props, () => {
...
+		padding-inline: 0 44px;
```
ShGKme, line 214: If this padding is defined by the a button size
```suggestion
		padding-inline: 0 var(--default-clickable-area);
```

src/views/ActivityAppFeed.vue
ShGKme: But we should also add a variable or component for that to not calculate it all the time in each app...

I'll do it later today.

susnux (changes requested):
This was done on purpose to make the list look like a list, as on bigger screens it becomes unreadable otherwise.
Same layout like e.g. on *forms* or others.

So the list is centered and fixed width. But I agree the scrollbar currently looks not good and should be moved to the very right instead directly next to the centered list.

With this changes it looks really weird and unreadable on wide screens, so I would request a review of @nextcloud/designers 

before | with this
---|---
Screenshot 2024-06-06 at 11-10-53 Aktivität - Aquarium|Screenshot 2024-06-06 at 11-11-46 Aktivität - Aquarium

solracsf:
> With this changes it looks really weird and unreadable on wide screens

Well, we think the same but with **current** implementation 😄 

I think the problem is that the date is floating right of the activity line. If they all (activity and date) were aligned left, i think this would be the best option.

Should this be validated, i could work on that if no one wants to do that.

But, a centered and fixed width doesn't seem the best option IMHO.

susnux:
> But, a centered and fixed width doesn't seem the best option IMHO.

That's why I cc the designers. We use the same layout also in Talk, Announcement Center, and Forms so basically all apps that are "lists" content wise

ShGKme:
> That's why I cc the designers. We use the same layout also in Talk, Announcement Center, and Forms so basically all apps that are "lists" content wise

Not really the same. Styles are local here and it looks different. For example, we don't have a scroll bar in the middle of content in Talk. It's on the corner.

ShGKme:
What about adding a light border between items, so it's easy to connect time and item visually?

src/views/ActivityAppFeed.vue
solracsf: No problem; same code a few lines down.

solracsf:
@ShGKme please feel free to take the lead on this one 👍

susnux:
> For example, we don't have a scroll bar in the middle of content in Talk. It's on the corner.

Yes as I wrote this is a bug currently but the limited width is on design purpose (see also current settings changes).


> the scrollbar currently looks not good and should be moved to the very right instead directly next to the centered list.

ShGKme:
@solracsf @susnux I've pushed a commit with an intermediate change:
- Scroll bar pushed to the page's edge
- The content is kept on the left (similar to most of the apps like Photos, Settings)
- Using more variables

But I'd make it even wide than in the past

Before | This PR | Proposal
---|---|---
image | image | image

marcoambrosini:
Width should be 900px. Your proposal looks good @ShGKme

ShGKme:
> If possible NcAppSettingsSection should be user

I'd not use `NcAppSettingsSection` even if we want a similar design... `NcAppSettingsSection` by definition has a different purpose. We should not think about activities and other usages while modifying settings components.

We can just use the same design here or introduce a new component.

marcoambrosini:
> We should not think about activities and other usages while modifying settings components

Many already do... Do you think that then it would make sense to create an ad hoc component there that all app developers can use?

ShGKme:
@marcoambrosini Do you mean `NcAppSettingsSection` or `NcSettingsSection`?

> Many already do...

Do you have an example?

jancborchardt (commented):
Fully agree with @susnux here, the scrollbar hanging next to the text is not by design and it should be all the way on the right.

@ShGKme your proposal looks good, but the content should still be visually centered. Especially on bigger screens the left-alignment looks strange. (Something we need to fix in Settings as well.)

marcoambrosini:
> Something we need to fix in Settings as well.

@jancborchardt settings should be left aligned as the sidebar is a integral part of the content and on a big screen they get too separated. That is unless we're able to center the navigation too

See the following screenshot as a reference. Granted: it's not great, but think about if the main content was centered
Screenshot 2024-08-19 at 15 47 43

solracsf:
> @ShGKme your proposal looks good, but the content should still be visually centered. Especially on bigger screens the left-alignment looks strange.

IMO, the centered content looks strange, as everything else is left aligned... 😄

ShGKme:
@jancborchardt @marcoambrosini 

This PR is open for quite a long time, and comments from the design team are different.

Are you against this open (settings like, limited width on the left with scroll bar on the right)?

image

marcoambrosini:
@ShGKme my comment was only concerning settings. I would prefer if there we kept the left alignment

jancborchardt:
@ShGKme so to clarify, for Activity:
- Limited width as before (can be wider, would be nice if it is the same as in Talk, Text etc)
- Centered as before
- Scrollbar on right edge

jancborchardt:
Additionally actually: The heading "All activities" (or the others) would also look better if it’s left aligned with the centered content, instead of all the way on the left.

solracsf:
> But I agree it would be nice for the scrollbar to be all the way on the right.

This PR is clearly not adpted for this.
Please close this one and take care of this from scratch.

PR #1723, opened 2024-07-11
fix(DigestSender): Do not add `and 1 more` in digest mail but simply include it

susnux:
1 more makes no sense as we can simply show that additional activity, so instead show it if there is only one more activity.

### DIFF
--- a/lib/DigestSender.php
+++ b/lib/DigestSender.php
@@ -120,20 +120,26 @@ class DigestSender {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):

miaulalala (approved):

artonge (approved):

susnux:
Cypress unrelated, can someone force merge?

PR #1730, opened 2024-07-16
fix access array offset in FilesHook.php

### LINKED ISSUE #1056: Trying to access array offset on value of type null at /lib/FilesHooks.php#471

Messj1:
prevent access array offset which doesn't exist and check if returned value is null

fixes #1056 

The Problem is related to the empty default array:
https://github.com/nextcloud/activity/blob/22bf3f2329e5c1e7cf2b9fa627a59c02ae8ae86a/lib/FilesHooks.php#L569-L580

compared to server implemented skeleton:
https://github.com/nextcloud/server/blob/d237fd0e78af8bbff51f2802efd4db1d88457579/lib/private/Share20/ShareHelper.php#L27-L32
```php
	public function getPathsForAccessList(Node $node) {
		$result = [
			'users' => [],
			'remotes' => [],
		];
```

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -242,6 +242,9 @@ class FilesHooks {
 			return;
 		}
 
+		// cache moveCase result until attributes are calculated for next stage (fileMovePost)
+		// moveCase has to be set as last part before return
+		$moveCase = $this->moveCase;
 		if (strpos($oldDir, $newDir) === 0) {
 			/**
 			 * a/b/c moved to a/c
@@ -251,7 +254,7 @@ class FilesHooks {
 			 * - a/b/ shared: delete
 			 * - a/ shared: move/rename
 			 */
-			$this->moveCase = 'moveUp';
+			$moveCase = 'moveUp';
 		} elseif (strpos($newDir, $oldDir) === 0) {
 			/**
 			 * a/b moved to a/c/b
@@ -261,7 +264,7 @@ class FilesHooks {
 			 * - a/c/ shared: add
 			 * - a/ shared: move/rename
 			 */
-			$this->moveCase = 'moveDown';
+			$moveCase = 'moveDown';
 		} else {
 			/**
 			 * a/b/c moved to a/d/c
@@ -272,7 +275,7 @@ class FilesHooks {
 			 * - a/d/ shared: add
 			 * - a/ shared: move/rename
 			 */
-			$this->moveCase = 'moveCross';
+			$moveCase = 'moveCross';
 		}
 
 		[$this->oldParentPath, $this->oldParentOwner, $this->oldParentId] = $this->getSourcePathAndOwner($oldDir);
@@ -291,6 +294,9 @@ class FilesHooks {
 		}
 
 		$this->oldAccessList = $oldAccessList;
+
+		// now its save to set moveCase
+		$this->moveCase = $moveCase;
 	}
 
 
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -578,6 +578,8 @@ class FilesHooks {
 		try {
 			$node = $this->rootFolder->getUserFolder($uidOwner)->get($path);
 		} catch (NotFoundException $e) {
+			$this->logger->warning('Path "{path}" with owner "{uidOwner}" was not found',
+				['path'=> $path, 'uidOwner'=>$uidOwner]);
 			return [
 				'users' => [],
 				'remotes' => [],
@@ -585,6 +587,8 @@ class FilesHooks {
 		}
 
 		if (!$node instanceof Node) {
+			$this->logger->warning('Path "{path}" of "{uidOwner}" is not a valid type.',
+				['path'=> $path, 'uidOwner'=>$uidOwner]);
 			return [
 				'users' => [],
 				'remotes' => [],
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -578,11 +578,17 @@ class FilesHooks {
 		try {
 			$node = $this->rootFolder->getUserFolder($uidOwner)->get($path);
 		} catch (NotFoundException $e) {
-			return [];
+			return [
+				'users' => [],
+				'remotes' => [],
+			];
 		}
 
 		if (!$node instanceof Node) {
-			return [];
+			return [
+				'users' => [],
+				'remotes' => [],
+			];
 		}
 
 		$accessList = $this->shareHelper->getPathsForAccessList($node);


lib/FilesHooks.php
```diff
@@ -396,7 +396,10 @@ protected function fileMoving($oldPath, $newPath) {
 		}
 		$accessList = $this->getUserPathsFromPath($parentPath, $parentOwner);
 		$affectedUsers = $accessList['users'];
-		$oldUsers = $this->oldAccessList['users'];
+        if(!isset($this->oldAccessList['users'])) return;
```
artonge, line 399:
```suggestion
	if (!isset($this->oldAccessList['users'])) {
		return;
	}
```

artonge: Is it really needed to return early? Aren't we losing the handling of potential remote users?

lib/FilesHooks.php
```diff
@@ -396,7 +396,10 @@ protected function fileMoving($oldPath, $newPath) {
 		}
 		$accessList = $this->getUserPathsFromPath($parentPath, $parentOwner);
 		$affectedUsers = $accessList['users'];
-		$oldUsers = $this->oldAccessList['users'];
+		$oldUsers = [];
+		if (isset($this->oldAccessList['users'])) {
+			$oldUsers = $this->oldAccessList['users'];
+		}
```
artonge, line 402: This should work:
```suggestion
		$oldUsers = $this->oldAccessList['users'] ?? [];
```

artonge (approved):

Messj1:
Think there could be a cleaner solution that fits the case:

Revert changes:
```diff
+		$oldUsers = $this->oldAccessList['users'];
-		$oldUsers = $this->oldAccessList['users'] ?? [];

...

+		$beforeRemotes = $this->oldAccessList['remotes'];
-		$beforeRemotes = $this->oldAccessList['remotes'] ?? [];
```

And set a default skeleton in case oldAccessList was not set (Error or Exception):

```diff
+		if($this->oldAccessList == null) {
+			$this->oldAccessList = [
+				'users' => [],
+				'remotes' => [],
+			];
+		}
		$oldUsers = $this->oldAccessList['users'];
```

So it would be simply to implement logging like:

```diff
		if($this->oldAccessList == null) {
+			$this->logger->warning("Something fishy happens: function fileMove was not finished");
			$this->oldAccessList = [
				'users' => [],
				'remotes' => [],
			];
		}
		$oldUsers = $this->oldAccessList['users'];
```

and 
```diff
	protected function getUserPathsFromPath($path, $uidOwner) {
		try {
			$node = $this->rootFolder->getUserFolder($uidOwner)->get($path);
		} catch (NotFoundException $e) {
+			$this->logger->warning("Something fishy happens: Path was not found", ['path'=> $path, 'uidOwner'=>$uidOwner]);
			return [
				'users' => [],
				'remotes' => [],
			];
		}

		if (!$node instanceof Node) {
+			$this->logger->warning("Something fishy happens: 'Folder' didn't return a Node", ['path'=> $path, 'uidOwner'=>$uidOwner]);
			return [
				'users' => [],
				'remotes' => [],
			];
		}
```

Opinions?

inspired by https://github.com/nextcloud/activity/pull/1281
> I am wondering if this won't juste hide another issue. Can you investigate ?

artonge:
Indeed, thanks for linking the existing PR. I forgot about that one.

Maybe having `getUserPathsFromPath` to return a skeleton instead of an empty array would be enough?
Adding logs is also a good idea to further address the issue later :).
Then, I don't think having a skeleton in `fileMoving` is necessary.

Messj1:
> Maybe having `getUserPathsFromPath` to return a skeleton instead of an empty array would be enough? 
> ...
> Then, I don't think having a skeleton in `fileMoving` is necessary.

This PR is related to 2 bugs:

1. The error handling in `getUserPathsFromPath` was not returning a proper value (skeleton)
2. If function `fileMove` is not called or if it not reach L277-L285

https://github.com/nextcloud/activity/blob/22bf3f2329e5c1e7cf2b9fa627a59c02ae8ae86a/lib/FilesHooks.php#L277-L285

So for the second case we need to check if `$this->oldAccessList` is not null

lib/FilesHooks.php
```diff
@@ -396,7 +396,16 @@ protected function fileMoving($oldPath, $newPath) {
 		}
 		$accessList = $this->getUserPathsFromPath($parentPath, $parentOwner);
 		$affectedUsers = $accessList['users'];
-		$oldUsers = $this->oldAccessList['users'];
+		$oldAccessList = $this->oldAccessList;
+		if($oldAccessList === null) {
+			$this->logger->warning("Something fishy happens: function fileMove was not finished: {oldPath} -> {path} : {uidOwner}",
```
artonge, line 401: As most returns from `fileMove` are legitimate and not fishy, I think I would not add a log here and rather add some in the `fileMove` move early returns that are unexpected. And I would also be more explicit in the message.
Sorry for the numerous back and forth, and thanks again for addressing the issue :).

Messj1: > As most returns from fileMove are legitimate and not fishy, I think I would not add a log here and rather add some in the fileMove move early returns that are unexpected. 

Yeah, good point.

The state `$oldAccessList === null` is illegal (unhealthy state). It means, that the `FilesHooks::fileMove`  function gets ended unexpected.


I count 4 return points in `FilesHooks::fileMove` :
1) https://github.com/nextcloud/activity/blob/70362c1c0d8d949d6af8f587a8e3409ea3c9265b/lib/FilesHooks.php#L218-L219
2) https://github.com/nextcloud/activity/blob/70362c1c0d8d949d6af8f587a8e3409ea3c9265b/lib/FilesHooks.php#L233-L234
3) https://github.com/nextcloud/activity/blob/70362c1c0d8d949d6af8f587a8e3409ea3c9265b/lib/FilesHooks.php#L273-L274
4) https://github.com/nextcloud/activity/blob/70362c1c0d8d949d6af8f587a8e3409ea3c9265b/lib/FilesHooks.php#L284-L286

And only the 4. case calls the `FilesHooks::fileMoving` function.

https://github.com/nextcloud/activity/blob/70362c1c0d8d949d6af8f587a8e3409ea3c9265b/lib/FilesHooks.php#L295-L312

A more appropriate way would be to invalidate the post process and let error be error, since it is generated by 3rd party.

Cache the case `$this->moveCase` instead of set it direct:

```diff
+		$this->moveCase = false;
+		$moveCase = false;
		if (strpos($oldDir, $newDir) === 0) {
			/**
			 * a/b/c moved to a/c
			 *
			 * Cases:
			 * - a/b/c shared: no visible change
			 * - a/b/ shared: delete
			 * - a/ shared: move/rename
			 */
-			$this->moveCase = 'moveUp';
+			$moveCase = 'moveUp';
		} elseif (strpos($newDir, $oldDir) === 0) {
			/**
			 * a/b moved to a/c/b
			 *
			 * Cases:
			 * - a/b shared: no visible change
			 * - a/c/ shared: add
			 * - a/ shared: move/rename
			 */
-			$this->moveCase = 'moveDown';
+			$moveCase = 'moveDown';
		} else {
			/**
			 * a/b/c moved to a/d/c
			 *
			 * Cases:
			 * - a/b/c shared: no visible change
			 * - a/b/ shared: delete
			 * - a/d/ shared: add
			 * - a/ shared: move/rename
			 */
-			$this->moveCase = 'moveCross';
+			$moveCase = 'moveCross';
		}
```

And set de case at the end to guarantee the availability of `$this->oldAccessList `

```diff
		}

+		$this->moveCase = $moveCase;
		$this->oldAccessList = $oldAccessList;
	}
```

So I would be able to remove the `null` check and logging because it is guaranteed

```diff
		$affectedUsers = $accessList['users'];
-		$oldAccessList = $this->oldAccessList;
-		if($oldAccessList === null) {
-			$this->logger->warning("Something fishy happens: function fileMove was not finished: {oldPath} -> {path} : {uidOwner}",
-				['oldPath'=>$oldPath, 'path'=> $parentPath, 'uidOwner'=>$parentOwner]);
-			$oldAccessList = [
-				'users' => [],
-				'remotes' => [],
-			];
-		}
-		$oldUsers = $oldAccessList['users'];
+		$oldUsers = $this->oldAccessList['users'];

	}
```

lib/FilesHooks.php
```diff
@@ -570,11 +572,21 @@ protected function getUserPathsFromPath($path, $uidOwner) {
 		try {
 			$node = $this->rootFolder->getUserFolder($uidOwner)->get($path);
 		} catch (NotFoundException $e) {
-			return [];
+			$this->logger->warning('Something fishy happens: Path "{path}" with owner "{uidOwner}" was not found',
```
artonge, line 575: It's a warning, no need to add such statement. :)
```suggestion
			$this->logger->warning('Path "{path}" with owner "{uidOwner}" was not found',
```

lib/FilesHooks.php
```diff
@@ -282,6 +283,7 @@ public function fileMove($oldPath, $newPath) {
 			$oldAccessList['users'] = array_merge($oldAccessList['users'], $this->getAffectedUsersFromCachedMounts($oldFileId));
 		}
 
+		$this->moveCase = $moveCase;
```
artonge, line 286: Could you add a comment here? I am not sure to understand why you are caching the value before assigning it?

lib/FilesHooks.php
```diff
@@ -283,6 +286,9 @@ public function fileMove($oldPath, $newPath) {
 		}
 
 		$this->oldAccessList = $oldAccessList;
+
+		// now its save to set moveCase
+		$this->moveCase = $moveCase;
```
artonge, line 291: I still don't get why this is safer to set it here

Messj1: There are 2 Signals invoked: 
https://github.com/nextcloud/activity/blob/06aabffe31252b54caf1fd439f65e48cf1971b95/lib/AppInfo/Application.php#L166-L167

The function `fileMove` is called before the action and the function `fileMovePost` afterwards.

`$this->moveCase ` gets only defined in function `fileMove`. 

In `fileMovePost` the `$this->moveCase` is consumed and will be reset

https://github.com/nextcloud/activity/blob/06aabffe31252b54caf1fd439f65e48cf1971b95/lib/FilesHooks.php#L310-L313

If the value of `$this->moveCase` is not set the function returns immediately.

https://github.com/nextcloud/activity/blob/06aabffe31252b54caf1fd439f65e48cf1971b95/lib/FilesHooks.php#L295-L299

Conclusion:
@artonge It is save the set the variable at the end cause everything **is** calculated  and there was no Error/Exception.
-> Ready to go to next stage

Currently it goes into post process and the `$this->oldAccessList` Variable is not defined because it gets defined at the end and in my case there was an exception thrown by 3rd party (circle) during calculate accesslist. :cry: 

https://github.com/nextcloud/activity/blob/06aabffe31252b54caf1fd439f65e48cf1971b95/lib/FilesHooks.php#L284-L286

come-nc: This should be fixed by catching Throwable in the try/catch and set moveCase to false in there before rethrowing, not by using a temporary var like this, this is too confusing.

PR #1744, opened 2024-08-01
fix(stream): Don't render avatar for federated users

nickvergessen:
Before | After
---|---
Bildschirmfoto vom 2024-08-01 12-23-24 | Bildschirmfoto vom 2024-08-01 12-21-12


src/utils/richObjects.ts
```diff
@@ -46,6 +46,17 @@ export function mapRichObjectToRichArgument(richObject: IRichObject) {
...
+			return {
+				component: NcUserBubble,
+				props: {
+					avatarImage: 'icon-user',
+					displayName: richObject.name,
+					user: richObject.id,
+					url: richObject.link,
+				},
+			}
+		}
```
Antreesy, line 59: Maybe replace existing code with it to avoid duplications?
```suggestion
		return {
			component: NcUserBubble,
			props: {
				// show thumbnail for federated users
				avatarImage: richObject.server ? 'icon-user' : undefined,
				displayName: richObject.name,
				user: richObject.id,
				url: richObject.link,
			},
		}
```

Antreesy (approved):
Otherwise fine

PR #1754, opened 2024-08-24
feat: Add sharing activity for teams

susnux:
This adds support for Teams (previously "Circles"), the user notification already work, but the notification for the owner and sharer need support in the `files_sharing` app.

So if this is merged we can add the required providers in `files_sharing`.

But one thing to mention: For groups we paginate the members, I did not find a nice way to do this with teams.
So maybe @ArtificialOwl can have a look at the member query if there is something to improve performance wise (e.g. pagination?).

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -11,6 +11,8 @@ use OC\Files\View;
 use OCA\Activity\BackgroundJob\RemoteActivity;
 use OCA\Activity\Extension\Files;
 use OCA\Activity\Extension\Files_Sharing;
+use OCA\Circles\CirclesManager;
+use OCA\Circles\Model\Member;
 use OCP\Activity\IManager;
 use OCP\Constants;
 use OCP\Files\Config\IUserMountCache;
@@ -60,7 +62,8 @@ class FilesHooks {
 		protected IUserMountCache $userMountCache,
 		protected IConfig $config,
 		protected NotificationGenerator $notificationGenerator,
-		protected ITagManager $tagManager
+		protected ITagManager $tagManager,
+		protected ?CirclesManager $teamManager,
 	) {
 	}
 
@@ -664,6 +667,16 @@ class FilesHooks {
 					(int)$share->getId()
 				);
 				break;
+			case IShare::TYPE_CIRCLE:
+				$this->shareWithTeam(
+					$share->getSharedWith(),
+					$share->getNodeId(),
+					$share->getNodeType(),
+					$share->getTarget(),
+					(int)$share->getId(),
+					$share->getSharedBy(),
+				);
+				break;
 			case IShare::TYPE_LINK:
 				$this->shareByLink(
 					$share->getNodeId(),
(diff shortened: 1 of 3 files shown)


miaulalala (approved):
LGTM

susnux:
TIL: `protected (Foo&Bar)|Baz $variable` does not work with PHP 8.1 but only 8.2 :sweat_smile:

lib/FilesHooks.php
```diff
@@ -726,7 +739,8 @@ protected function shareWithGroup($shareWith, $fileSource, $itemType, $fileTarge
 		$offset = 0;
 		$users = $group->searchUsers('', self::USER_BATCH_SIZE, $offset);
 		while (!empty($users)) {
-			$this->addNotificationsForGroupUsers($users, 'shared_with_by', $fileSource, $itemType, $fileTarget, $shareId);
+			$users = array_map(fn (IUser $user) => $user->getUID(), $users);
```
artonge, line 742:
```suggestion
			$usersIds = array_map(fn (IUser $user) => $user->getUID(), $users);
```

lib/FilesHooks.php
```diff
@@ -758,6 +772,42 @@ protected function shareByLink($fileSource, $itemType, $linkOwner) {
...
+		if ($this->teamManager === null) {
+			return;
+		}
+
+		try {
+			$this->teamManager->startSuperSession();
+			$team = $this->teamManager->getCircle($shareWith);
+			$members = $team->getInheritedMembers();
+			$members = array_filter($members, fn ($member) => $member->getUserType() === Member::TYPE_USER);
+			$users = array_map(fn ($member) => $member->getUserId(), $members);
```
artonge, line 794:
```suggestion
			$usersIds = array_map(fn ($member) => $member->getUserId(), $members);
```

lib/FilesHooks.php
```diff
@@ -758,6 +772,42 @@ protected function shareByLink($fileSource, $itemType, $linkOwner) {
...
+	 * @param string $fileTarget File path
+	 * @param int $shareId The Share ID of this share
+	 */
+	protected function shareWithTeam(string $shareWith, int $fileSource, string $itemType, string $fileTarget, int $shareId, string $sharer): void {
+		if ($this->teamManager === null) {
+			return;
+		}
+
+		try {
+			$this->teamManager->startSuperSession();
```
artonge, line 790: What does it do? Does it needs to be stoped?

susnux: Start the session. It allows to query also users that could not be detected by current user due to access restrictions (we want to inform all members so we need a session without a current user set).

https://github.com/nextcloud/circles/blob/08a08b12e2ca8207a2d8fa09a410363ace9b4719/lib/CirclesManager.php#L168

susnux: (I never saw that the session is stopped in collectives or other apps that use teams)

ArtificialOwl (approved):

lib/FilesHooks.php
```diff
@@ -878,9 +928,11 @@ protected function unshareFromGroup(IShare $share) {
 
 		$offset = 0;
 		$users = $group->searchUsers('', self::USER_BATCH_SIZE, $offset);
+		$users = array_map(fn (IUser $user) => $user->getUID(), $users);
 		$shouldFlush = $this->startActivityTransaction();
 		while (!empty($users)) {
-			$this->addNotificationsForGroupUsers($users, $actionUser, $share->getNodeId(), $share->getNodeType(), $share->getTarget(), (int)$share->getId());
+			$userIds = array_map(fn (IUser $user) => $user->getUID(), $users);
```
come-nc, line 934: $users contains string[] after line 931, then you try to convert to string again.
Results in:
```
Argument #1 ($user) must be of type OCP\IUser, string given in file 'server/server/apps/activity/lib/FilesHooks.php' line 935
```

susnux: (or remove 934 to keep the activity transaction as short as possible)

PR #1776, opened 2024-09-13
notice on empty cachedPath

ArtificialOwl:
notice and ignore entries with empty cachedPath for userId.

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -1384,6 +1384,12 @@ class FilesHooks {
 		// now that we have a list of eventuals filtered users, we confirm they have no access to the file
 		$filteredUsers = [];
 		foreach ($usersToCheck as $userId) {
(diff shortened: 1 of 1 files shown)


lib/FilesHooks.php
```diff
@@ -1331,6 +1331,12 @@ private function getUnrelatedUsers(int $fileId, array $cachedMounts): array {
 		// now that we have a list of eventuals filtered users, we confirm they have no access to the file
 		$filteredUsers = [];
 		foreach ($usersToCheck as $userId) {
+			if (!array_key_exists($userId, $cachedPath) || $cachedPath[$userId] === null) {
+				$this->logger->notice('could not find user in list of cached path', ['cachePath' => $cachedPath, 'usersToCheck' => $usersToCheck, 'current' => $userId]);
```
artonge, line 1335: I don't know whether it is an error or not. Shouldn't we try to fetch the user path?

PR #1785, opened 2024-09-14
chore: Use composer-bin-plugin to manage dev deps

artonge:
As https://github.com/nextcloud/photos/pull/2629


come-nc (approved):
phpunit should go in its own folder as well

PR #1803, opened 2024-09-24
fix: do not send daily digest email to user who is disabled

yemkareems:
Add a extra check to not send the email to user who is disabled.

In my local i made jane a disabled user 

image

And add the condition. This triggered a email to alice, bob and john. And it did not send the email to jane

image


artonge (approved):
Seems enough :+1:

come-nc (commented):
Would it be possible to instead take an IUser object in sendDigestForUser and check the enabled state in parent method sendDigests ?
That would allow to also not update the activity.digest config value for disabled users, as that may be confusing. Would also be a bit better in performance to bail out earlier for disabled users.

provokateurin (approved):

nickvergessen:
> That would allow to also not update the activity.digest config value for disabled users, as that may be confusing. Would also be a bit better in performance to bail out earlier for disabled users.

I was wondering if we should even call
```php
$this->config->setUserValue($user->getUID(), 'activity', 'activity_digest_last_send', (string)$lastActivityId);
```
so after re-enabling the user is not spamed with old legacy info from the day they got disabled.

yemkareems:
> > That would allow to also not update the activity.digest config value for disabled users, as that may be confusing. Would also be a bit better in performance to bail out earlier for disabled users.
> 
> I was wondering if we should even call
> 
> ```
> $this->config->setUserValue($user->getUID(), 'activity', 'activity_digest_last_send', (string)$lastActivityId);
> ```
> 
> so after re-enabling the user is not spamed with old legacy info from the day they got disabled.

Hi @nickvergessen , @come-nc Do we need to do this for disabled user. I see there is a ACTIVITY_LIMIT and first time after re enabling user the notifications ll be cluttered and after that it should be fine. Please let me know and i ll do it accordingly

yemkareems:
Hi @come-nc , @nickvergessen I will add the user disabled check in sendDigests method as come suggested but will also add one updateLastSentForUser which will do 
$this->config->setUserValue($user->getUID(), 'activity', 'activity_digest_last_send', (string)$lastActivityId);
as Joas suggested.

yemkareems:
Hi @come-nc, @nickvergessen I have done this change as requested but however i do not see a entry in oc_preferences for the disabled user(jane). I see the mail has gone out for admin user and only entry for admin user and not for the disabled user jane for whom i did some file rename just before disabling.

image

nickvergessen:
Could you push, so we can see what your code is?

miaulalala (approved):

lib/DigestSender.php
```diff
@@ -68,6 +69,12 @@ public function sendDigests(int $now): void {
 				// User got todays digest already
 				continue;
 			}
+			$user = $this->userManager->get($user);
```
miaulalala, line 72: You will need to change Line 88 to this as you're overwriting the user

```
$this->config->setUserValue($user->getUID(), 'activity', 'digest', $timezoneDigestDay[$timezone]);
```

miaulalala (changes requested):

lib/DigestSender.php
yemkareems: Oops sorry about it. Fixed the same.

lib/DigestSender.php
```diff
@@ -103,11 +110,21 @@ private function getLastSendActivity(string $user, int $now): int {
 		return $this->data->getFirstActivitySince($user, $now - (24 * 60 * 60));
 	}
 
-	public function sendDigestForUser(string $uid, int $now, string $timezone, string $language) {
+	public function updateLastSentForUser(IUser $user, int $now): void {
```
come-nc, line 113:
```suggestion
	private function updateLastSentForUser(IUser $user, int $now): void {
```

lib/DigestSender.php
```diff
@@ -103,11 +110,21 @@ private function getLastSendActivity(string $user, int $now): int {
...
+		$uid = $user->getUID();
+		$lastSend = $this->getLastSendActivity($uid, $now);
+
+		['count' => $count, 'max' => $lastActivityId] = $this->data->getActivitySince($uid, $lastSend, true);
```
come-nc, line 117:
```suggestion
		['max' => $lastActivityId] = $this->data->getActivitySince($uid, $lastSend, true);
```
`$count` is not used

come-nc (commented):
Good apart from 2 small changes.

artonge (approved):

PR #1806, opened 2024-10-07
fix: unify file flags

stefanschramm:
`find . -executable -type f | grep -v "^./.git/hooks"` reveals some files that have the executable bit set. I guess this is not necessary.


artonge (approved):

artonge:
@stefanschramm can you fix DCO?

artonge:
Thanks @stefanschramm :)

PR #1705, opened 2024-06-09
[stable28] Fix npm audit

nextcloud-command:
# Audit report

This audit fix resolves 22 of the total 31 vulnerabilities found in your project.

## Updated dependencies
* [@nextcloud/dialogs](#user-content-\@nextcloud\/dialogs)
* [@nextcloud/files](#user-content-\@nextcloud\/files)
* [@nextcloud/moment](#user-content-\@nextcloud\/moment)
* [@nextcloud/vite-config](#user-content-\@nextcloud\/vite-config)
* [@testing-library/vue](#user-content-\@testing-library\/vue)
* [@vitejs/plugin-vue2](#user-content-\@vitejs\/plugin-vue2)
* [@vue/language-core](#user-content-\@vue\/language-core)
* [@vue/test-utils](#user-content-\@vue\/test-utils)
* [axios](#user-content-axios)
* [braces](#user-content-braces)
* [cross-spawn](#user-content-cross-spawn)
* [elliptic](#user-content-elliptic)
* [happy-dom](#user-content-happy-dom)
* [micromatch](#user-content-micromatch)
* [nanoid](#user-content-nanoid)
* [rollup](#user-content-rollup)
* [vite](#user-content-vite)
* [vite-plugin-dts](#user-content-vite-plugin-dts)
* [vue-resize](#user-content-vue-resize)
* [vue-template-compiler](#user-content-vue-template-compiler)
* [vue-tsc](#user-content-vue-tsc)
* [vuex](#user-content-vuex)
### @nextcloud/dialogs <a href="#user-content-\@nextcloud\/dialogs" id="\@nextcloud\/dialogs">#</a>
* Caused by vulnerable dependency:
  * [@nextcloud/files](#user-content-\@nextcloud\/files)
  * [@nextcloud/l10n](#user-content-\@nextcloud\/l10n)
  * [@nextcloud/vue](#user-content-\@nextcloud\/vue)
  * [vue](#user-content-vue)
  * [vue-frag](#user-content-vue-frag)
* Affected versions: >=2.0.0
* Package usage:
  * `node_modules/@nextcloud/dialogs`

### @nextcloud/files <a href="#user-content-\@nextcloud\/files" id="\@nextcloud\/files">#</a>
* Caused by vulnerable dependency:
  * [@nextcloud/l10n](#user-content-\@nextcloud\/l10n)
* Affected versions: >=1.1.0
* Package usage:
  * `node_modules/@nextcloud/files`

[...]

PR #1842, opened 2024-11-15
Revert "[stable28] fix/remote activity constructor"

juliusknorr:
Reverts nextcloud/activity#1813 as the code in 28 does not have the original issue.

The bug was only introduced in https://github.com/nextcloud/activity/commit/c8248b0114ea53c19c4298714c51090ae3252b15 for 29 and later


GretaD (approved):

juliusknorr:
php-cs also fails on other prs, seems unrelated (https://github.com/nextcloud/activity/pull/1705)

Issue #1850, opened 2024-11-23
[Bug] Incorrect margins in app heading for RTL mode


falghamdi125:
Hello everyone,

Nextcloud's RTL mode is currently in development and will be introduced in release 31.

When you switch to RTL mode by selecting an RTL language like Arabic, you'll notice an issue with the activity app title margins. It's currently optimized for LTR layouts, which causes it to overlap with the navigation sidebar toggle button. For proper RTL support, the title margin should be adjusted slightly to the left to ensure the toggle button fits neatly in the available space.

![Image](https://github.com/user-attachments/assets/7bc8143e-9d52-4f00-8424-faeb19d05519)

---

### Steps to reproduce
1.  Ensure you are running Nextcloud, Activity and Nextcloud-vue libs master branches, 
2.  Switch to Arabic
3.  Navigate to the Activity app

### Expected behaviour
App title is moved to the left a little bit to allow the toggle button to be displayed.

### Actual behaviour
Overlaping between the toggle button and the app title.

### Server configuration

**Nextcloud version:** 
master

**nextcloud-vue libraries:**
master

PR #1851, opened 2024-11-23
fix: Add RTL support to app heading.

falghamdi125:
### ☑️ Resolves  

-  #1850 

Hello everyone,

This PR adjusts app heading margins and adds support for RTL layouts.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
| -- | -- |
| image |  image |

💡   Note: This pull request is most effective when the Nextcloud-Vue  [PR:6260]( https://github.com/nextcloud-libraries/nextcloud-vue/pull/6260)  is approved.

Best


artonge (approved):

src/views/ActivityAppFeed.vue
```diff
@@ -222,7 +222,7 @@ watch(props, () => {
 		font-size: 20px;
 		line-height: 44px; // to align height with the app navigation toggle
 		// Align with app navigation toggle
-		margin: var(--app-navigation-padding, 8px) 0 0 calc(2 * var(--app-navigation-padding, 8px) + 44px);
+		margin-inline: calc(2 * var(--app-navigation-padding, 8px) + 44px) var(--app-navigation-padding, 8px);
```
artonge, line 225: I understand that previously, `margin-top` and `margin-left` were set, and now that `margin-inline-start` `margin-inline-end` are set, which are equivalent of `margin-left` and `margin-right`. Am I right? If so, don't we need to keep the `margin-top`?

falghamdi125: I am not sure about the top margin.

Keeping the top margin is ok but it is not aligned with the toggle button. 

image

Setting the top margin to zero aligns the title with the toggle button. See:
image

Probably the designers are looking for comfortable margins around app content more than alignment with external elements. It guess it is safe to maintain the top margin.

artonge: Your second screenshot looks good indeed. Let's keep it that way :).

falghamdi125:
Build files have been added to the commit @artonge

artonge:
Thanks a lot @falghamdi125 !

We have a Nextcloud Talk instance where we invite Nextcloud contributors to be able to communicate more quickly. If you like, I could give you a guest account via the email address in your commits?

PR #1916, opened 2025-03-17
 fix: update workflows and remove default_enabled from info.xml

miaulalala:


.github/workflows/cypress.yml
```diff
@@ -95,13 +95,13 @@ jobs:
       matrix:
         # Run multiple copies of the current job in parallel
         # Please increase the number or runners as your tests suite grows
-        containers: ['1', '2', '3']
+        containers: ['component', '1', '2', '3']
```
miaulalala, line 98: @nickvergessen is this correct?

artonge: You can drop it, else Cypress will complain.

appinfo/info.xml
```diff
@@ -21,7 +21,6 @@
 	<author>Joas Schilling</author>
 	<author>Ferdinand Thiessen</author>
 
-	<default_enable />
```
artonge, line 24: I think this is needed

nickvergessen (approved):

PR #1926, opened 2025-03-18
chore: do not use private API (`OC`) but instead use `OCP`

susnux:


miaulalala:
I guess the baseline needs updating?

come-nc (approved):
nextcloud/rector can do it for you

PR #1927, opened 2025-03-18
feat: Use X-NC-Nickname as user identifier

artonge:


lib/CurrentUser.php
```diff
@@ -10,28 +10,22 @@
...
 
-	/** @var string */
-	protected $identifier;
-	/** @var string|null */
-	protected $cloudId;
-	/** @var string|false|null */
-	protected $sessionUser;
+	protected string|null $identifier;
+	protected string|null $cloudId;
+	protected string|false|null $sessionUser;
```
miaulalala, line 22:
```suggestion
	protected ?string $identifier;
	protected ?string $cloudId;
	protected ?string $sessionUser;
```

Can we avoid the triple type for the `$sessionUser` somehow? Haven't checked all returns but it would be nice if we could get rid of the `false`.

lib/CurrentUser.php
```diff
@@ -10,28 +10,22 @@
...
-	 * @param IRequest $request
-	 * @param IManager $shareManager
-	 */
 	public function __construct(
 		protected IUserSession $userSession,
 		protected IRequest $request,
 		protected IManager $shareManager,
+		protected IFactory $l10nFactory,
 	) {
 		$this->cloudId = false;
```
miaulalala, line 30:
```suggestion
		$this->cloudId = null;
```

lib/CurrentUser.php
```diff
@@ -46,19 +40,30 @@ public function getUser(): ?IUser {
...
+			$this->identifier = $cloudId;
+			return $this->identifier;
+		}
 
-				if ($this->identifier === null) {
-					// Nothing worked, fallback to empty string
-					$this->identifier = '';
-				}
-			}
+		$nickname = $this->request->getHeader('X-NC-Nickname');
```
miaulalala, line 59: We should validate this.

lib/CurrentUser.php
```diff
@@ -46,19 +40,30 @@ public function getUser(): ?IUser {
...
+		if ($nickname !== '') {
+			$this->identifier = $nickname . ' (' . $this->l10nFactory->get('comments')->t('remote user') . ')';
```
miaulalala, line 61:
```suggestion
			$this->identifier = $nickname . ' (' . $this->l10nFactory->get('comments')->t('guest') . ')';
```

miaulalala (changes requested):

lib/CurrentUser.php
miaulalala: Same for the `$sessionUser`

artonge: Actually no, as this contains a cached value, and false is used to distinguished whether the cache is initialized or not.

lib/CurrentUser.php
artonge: See other comment, we need the false value. Even for $cloudId

lib/CurrentUser.php
artonge: Actually, not sure that we need any validation at all :thinking:

miaulalala: Idk, where does this string end up and is it always escaped properly? If not, I'd at least remove special chars

artonge: It will be displayed by vue, so safe, and stored in the DB, safe too.
And probably displayed in the activity mails. Might be an issue. I'll remove the special chars :+1:

PR #1931, opened 2025-03-20
fix: disable loading when there are no more activities to load

### LINKED ISSUE #1909: Calendar activities "endless spinner"
Endless spinner when selecting the calendar activities.

Details will be added later

Image

hamza221:
Fix  #1909
I couldn't reproduce in a deterministic manner `useInfiniteScroll` doesn't always trigger for pages with 1,2 activities disabling it makes reproduction easier. 

Repro:
1. Have a tab with one activity 
2. If the api request doesn't get triggered a second time with a `304`response loading Icons doesn't disappear


artonge (approved):

src/views/ActivityAppFeed.vue
```diff
@@ -156,7 +156,10 @@ async function loadActivities() {
 		)
 		allActivities.value.push(...response.data.ocs.data.map((raw) => new ActivityModel(raw)))
 		lastActivityLoaded.value = response.headers['x-activity-last-given']
-		hasMoreActivites.value = true
+		// If less than the hardcoded limit, there are no more activities
+		if (response.data.ocs.data.length < 50) {
+			hasMoreActivites.value = false
```
susnux, line 161: Possible improvement in the future: Check again after a timeout if there is something new.

susnux (approved):

src/views/ActivityAppFeed.vue
nickvergessen: This completely breaks the activity stream and needs reverting!

Basically if you upload more than 50 files, it will be the end of your activity stream. In fact 3 uploads will be enough already. You ask the backend for 50 entries, but only get 48, as 3 of them are combined into 1 entry

I see it's also broken before with the loading spinner you tried to fix, but that is another problem. Checking the result size is not the solution. Whether or not you reached the end of the activity stream is indicated by the HTTP status code 304:
https://github.com/nextcloud/activity/blob/master/docs/endpoint-v2.md#http-status

Issue #1889, opened 2025-02-27
Nextcloud email notification not sent when files are uploaded or deleted to shares


ETLTinhThanh:
The problem is that an email notification is not sent when files are uploaded or deleted to shares (no groupfolder involved) in Nextcloud.
Email notifications are only sent when files are modified or moved.

### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Steps to reproduce
1.
2.
3.

### Expected behaviour
Tell us what should happen

### Actual behaviour
Tell us what happens instead, if possible also add a screenshot

### Server configuration

**Web server:** Apache/Nginx

**Database:** MySQL/Maria/SQLite/PostgreSQL

**PHP version:** 8.1/8.2/8.3

**Nextcloud version:** (see Nextcloud admin page)

<details>
<summary>List of activated apps</summary>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```
</details>

<details>
<summary>Nextcloud configuration</summary>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder
```
</details>

### Browser

**Browser name:** Firefox/Chrome/Safari/…

**Browser version:** 124/125/…

**Operating system:** Windows/Ubuntu/Mac/…

<details>
<summary>Browser log</summary>

```
Insert your browser log here, this could for example include:
a) The javascript console log
b) The network log
c) ...
```

</details>

LaSepp:
There is already a discussion over at the [nextcloud / server issues](https://github.com/nextcloud/server/issues/49133) with a workaround.
Let's hope a final fix finds its way into the new NC versions

raph146:
I still encounter this issue in Nextcloud 34.0.1 with a folder shared via public link : even with the "Files have been uploaded to a folder shared by mail or by public link" notifications on and a working SMTP config nothing is sent.
Other notifications are working fine, any idea on how to fix this ?

PR #1947, opened 2025-03-26
fix(settings): ensure we handle file_created, file_deleted and file_restored as file_changed

skjnldsv:
The provider is `file_changed`, but we still have some related events. But they are still covered under one setting provider: https://github.com/nextcloud/server/blob/af6de04e9e141466dc229e444ff3f146f4a34765/apps/files/lib/Activity/Settings/FileChanged.php

image

Fix https://github.com/nextcloud/server/issues/49133
Fix https://github.com/nextcloud/activity/issues/1889

### DIFF
--- a/lib/UserSettings.php
+++ b/lib/UserSettings.php
@@ -7,6 +7,7 @@
 
 namespace OCA\Activity;
 
+use OCA\Activity\Extension\Files;
 use OCP\Activity\ActivitySettings;
 use OCP\Activity\Exceptions\SettingNotFoundException;
 use OCP\Activity\IManager;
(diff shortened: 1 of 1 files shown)


skjnldsv:
Some after thoughts, could be a clean way to add child events to the [`ActivitySettings`](https://github.com/nextcloud/server/blob/af6de04e9e141466dc229e444ff3f146f4a34765/lib/public/Activity/ActivitySettings.php). So we know we can group them without having to hardcode them into activity.
As a fix this is fine, you can see the use case was hardcoded 4 years ago already.
But for 32, it could be a nice way. Not sure if other providers would benefit from it

provokateurin (approved):

skjnldsv:
Seems like other PRs on master are failing with that cypress test, merging as unrelated

PR #1975, opened 2025-04-09
fix: Improving logging of invalid parameters for activity subject and…

nickvergessen:
… message

### DIFF
--- a/lib/GroupHelper.php
+++ b/lib/GroupHelper.php
@@ -61,6 +61,10 @@ class GroupHelper {
 				} else {
 					$event = $provider->parse($language, $event);
 				}
+
+				if ($event->isValidParsed()) {
+					$this->logger->info('Activity event was claimed to be parsed, but was not fully parsed by ' . get_class($provider) . ' [app: ' . $event->getApp() . ', subject: ' . $event->getSubject() . ']');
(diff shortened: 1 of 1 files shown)


lib/GroupHelper.php
```diff
@@ -61,6 +61,10 @@ public function addEvent(int $id, IEvent $event): void {
 				} else {
 					$event = $provider->parse($language, $event);
 				}
+
+				if ($event->isValidParsed()) {
+					$this->logger->info('Activity event was claimed to be parsed, but was not fully parsed by ' . get_class($provider) . ' [app: ' . $event->getApp() . ', subject: ' . $event->getSubject() . ']');
```
ChristophWurst, line 66: consider

```suggestion
					$this->logger->info('Activity event was claimed to be parsed, but was not fully parsed by ' . get_class($provider) . ' [app: ' . $event->getApp() . ', subject: ' . $event->getSubject() . ']', ['app' => $event->getApp()]);
```

so logs and Sentry point to the problematic app and not the activity app. same below.

nickvergessen: Makes sense, done

miaulalala (approved):

PR #2041, opened 2025-05-28
fix(notification): Deduplicate the settings check

nickvergessen:
This also fixes a problem with the files app that has a wrong setting type at this place otherwise as the type is overwritten after checking the settings.


nickvergessen:
@maintainers please check how far this needs backporting

PR #2052, opened 2025-06-04
fix(FilesHooks): Catch all exceptions when looking up unrelated users

artonge:
This will catch "Undefined array key" exceptions from above `$cachedPath[$userId]`.

Fix: https://github.com/nextcloud/groupfolders/issues/3680

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -1398,7 +1398,7 @@ class FilesHooks {
 				if ($node->isReadable()) {
 					continue; // overkill ? as rootFolder->get() would throw an exception if file is not available
 				}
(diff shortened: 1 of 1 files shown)


szaimen:
!/backport to stable28

artonge:
> should we log something ?

Not sure, as at that point, we do not know why the user was is not in `$cachedPath`

Best would have been to add an `if (isset($cachedPath[$userId]))` above, and to catch-ignore specific exceptions, and to catch-log the others. But not sure what exceptions are supposed to be thrown here.

Considering that I don't know enough to change the code further, and that the initial intention of the code was probably to catch all exceptions, I would stick with the current solution.

PR #2078, opened 2025-07-07
feat: use full date and time in activity emails

miaulalala:
Fixes https://github.com/nextcloud/activity/issues/2076

Also refactors the query for affected users to avoid using a `where` overwrite for the query.


nickvergessen (approved):
But otherwise fine by me

PR #2098, opened 2025-08-11
perf: pass the node of shares around instead getting the path from id each time

icewind1991:
No need for a expensive by-id searches if the share event already has the node


icewind1991:
The entire `FileHooks` is a mess performance wise...

CarlSchwan (approved):

PR #2135, opened 2025-09-18
fix(notification): missing notification

luka-nextcloud:
* Resolves: https://github.com/nextcloud/deck/issues/6899


nickvergessen:
This feels wrong/unexpected to change like that after 5 years of working like that?
The activity app takes care of parsing those notifications. If you don't want that, disable the option to receive activities as notifications for your type and handle the notifications in your app directly?

For now a similar result should be achievable by changing:
https://github.com/nextcloud/deck/blob/main/lib/Notification/Notifier.php#L80

```php
if ($notification->getApp() !== 'deck' || $notification->getObjectType() === 'activity_notification') {
```

luka-nextcloud:
Now I understand what you mean, thank you.

Issue #2148, opened 2025-09-29
Error deleting the share Nextcloud 31.0.9


USemyon:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Steps to reproduce
1. Create Folder
2. Share Folder with 2 internal user
3. Give one of them rights "Edit" or "Custom permissions "read, write, share"
4. Share it
5. As internal user with "share" permission open "sharing" settings of that folder
6. Edit Share -> Delete share of second user
7. You got the error 

### Expected behaviour
Share is being deleted

### Server configuration

**Web server:** Apache

**Database:** Maria

**PHP version:** 8.2

**Nextcloud version:** 31.0.9

<details>
<summary>List of activated apps</summary>

Enabled:
  - activity: 4.0.0
  - admin_audit: 1.21.0
  - cloud_federation_api: 1.14.0
  - comments: 1.21.0
  - dashboard: 7.11.0
  - dav: 1.33.0
  - deck: 1.15.3
  - federatedfilesharing: 1.21.0
  - federation: 1.21.0
  - files: 2.3.1
  - files_pdfviewer: 4.0.0
  - files_sharing: 1.23.1
  - files_trashbin: 1.21.0
  - files_versions: 1.24.0
  - firstrunwizard: 4.0.0
  - logreader: 4.0.0
  - lookup_server_connector: 1.19.0
  - nextcloud_announcements: 3.0.0
  - notifications: 4.0.0
  - oauth2: 1.19.1
  - password_policy: 3.0.0
  - privacy: 3.0.0
  - profile: 1.0.0
  - provisioning_api: 1.21.0
  - quota_warning: 1.22.0
  - recommendations: 4.0.0
  - richdocuments: 8.7.5
  - serverinfo: 3.0.0
  - settings: 1.14.0
  - survey_client: 3.0.0
  - text: 5.0.0
  - theming: 2.6.1
  - twofactor_backupcodes: 1.20.0
  - updatenotification: 1.21.0
  - user_status: 1.11.0
  - viewer: 4.0.0
  - workflowengine: 2.13.0
Disabled:
  - app_api: 5.0.2 (installed 5.0.2)
[...]

PR #2217, opened 2025-11-10
[stable31] fix: catch NotFoundException when user is resharer

miaulalala:
Since the file source path and user path for resharing users doesn't match, catch the exception and don't generate a notification instead.

Issue might be from https://github.com/nextcloud/activity/pull/2103, but the exception is also not bubbling up for 32+, so something else must have changed as well.

Fixes https://github.com/nextcloud/activity/issues/2148


miaulalala:
> Btw, it's still feels weirf that we cannot compute the path properly. The end user should still get the notification, no ?

We don't notify for 32+ either, it just fails more gracefully. I guess we could add a check on the file owner and send them a notification instead, but then we'll have to backport from master.

PR #2514, opened 2026-04-14
fix(notifications): Require absolute links for support of desktop and mobile clients

miaulalala:
To align with the stricter validation introduced in nextcloud/server#59606.

AI-Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/lib/NotificationGenerator.php
+++ b/lib/NotificationGenerator.php
@@ -31,6 +31,14 @@ class NotificationGenerator implements INotifier {
 	) {
 	}
 
+	private function sanitizeUrl(string $url): string {
+		if (str_starts_with($url, 'http://') || str_starts_with($url, 'https://')) {
+			return $url;
+		}
+
+		return '';
+	}
+
 	public function deferNotifications(): bool {
 		return $this->notificationManager->defer();
 	}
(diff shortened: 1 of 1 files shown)


lib/NotificationGenerator.php
```diff
@@ -28,9 +29,24 @@ public function __construct(
...
 
+	private function sanitizeUrl(string $url): string {
+		if (str_starts_with($url, 'http://') || str_starts_with($url, 'https://')) {
+			return $url;
+		}
+
+		// Reject non-HTTP(S) schemes (e.g. javascript:, nc://, data:)
+		if (preg_match('/^[a-zA-Z][a-zA-Z0-9+\-.]*:/', $url)) {
+			return '';
+		}
```
artonge: Can this really happen?

artonge: I would drop it if it is not needed.

miaulalala: It can, nc:// for example

nickvergessen: I would simply not set a link in call cases not covered by the first if.
That's also how notifications handles it.

PR #2548, opened 2026-05-02
feat(digest): inline file thumbnails in the daily activity digest email

karlitschek:
The daily activity digest currently lists each event with the provider's small icon URL. For file-related events that's a generic mime-type icon, which makes the digest feel like a server log rather than a "things that happened to your stuff today" summary. Show the actual file preview instead, when one is available.

`DigestSender::getPreviewDataUri()` resolves the file as the affected user (so access mirrors what the recipient sees in the web UI), generates a 96×96 preview via `IPreview`, and returns a base64 data URI inlined directly into the email — no public preview endpoint needed, and the digest renders fully offline.

Returns null on any failure (not a file event, file deleted, no read permission, preview generation unsupported, anything thrown). The existing `$event->getIcon()` is the seamless fallback, so existing behaviour is preserved for non-file or no-preview events.

The 96px cap keeps each thumbnail around 3-5 KB; even at the 20-event ACTIVITY_LIMIT the digest stays well under common message-size thresholds.

`IRootFolder` and `IPreview` are auto-wired through standard DI; `DigestSender` is not explicitly registered in `Application.php`, so no AppInfo change is needed.

### DIFF
--- a/lib/DigestSender.php
+++ b/lib/DigestSender.php
@@ -11,8 +11,13 @@ namespace OCA\Activity;
 use OCP\Activity\IEvent;
 use OCP\Activity\IManager;
 use OCP\Defaults;
+use OCP\Files\File;
+use OCP\Files\IRootFolder;
+use OCP\Files\NotFoundException;
+use OCP\Files\NotPermittedException;
 use OCP\IConfig;
 use OCP\IDateTimeFormatter;
+use OCP\IPreview;
 use OCP\IURLGenerator;
 use OCP\IUser;
 use OCP\IUserManager;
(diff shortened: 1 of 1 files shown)


lib/DigestSender.php
```diff
@@ -215,6 +231,50 @@ public function sendDigestForUser(IUser $user, int $now, string $timezone, strin
...
+			if (!$this->previewManager->isAvailable($node)) {
+				return null;
+			}
+
+			$preview = $this->previewManager->getPreview($node, self::PREVIEW_PIXELS, self::PREVIEW_PIXELS, true);
+			$content = $preview->getContent();
+			if ($content === '') {
+				return null;
+			}
+			$mime = $preview->getMimeType() ?: 'image/png';
```
miaulalala, line 264: This does image processing for every preview - and may be quite heavy for documents such as pdfs and office documents.

miaulalala (changes requested):

PR #2549, opened 2026-05-02
feat(admin): add database panel with size on disk and retention sugge…

karlitschek:
…stion

The activity log can grow into hundreds of GB on busy installations without the admin noticing — there is no built-in signal that something has gotten out of hand. Surface the relevant numbers in the Activity admin section.

New `DatabaseStats` service queries on-disk size against the activity-scoped connection (the dedicated one if `activity_db*` is configured, otherwise the main DB):

- MySQL/MariaDB: `information_schema.tables` (data_length + index_length).
- PostgreSQL: `pg_total_relation_size()` (table + indexes + toast).
- SQLite: returns null per table; admin UI explains the limitation.

The same service produces a conservative retention suggestion — when the total size crosses 1/5/10 GB and `activity_expire_days` is still at the 365-day default, recommend lowering it to 180/90/30 days. Recommendations only ever shorten; admins who already tuned retention down are not second-guessed. Suggestions are surfaced as a warning card in the new "Database" section of the admin UI; nothing is auto-applied — admins copy the `'activity_expire_days' => N,` snippet into `config.php` themselves so config-as-code workflows stay in control.

Wired through:
- `lib/DatabaseStats.php` — new, uses `IQueryBuilder::PARAM_STR_ARRAY`, honours `activity_dbtableprefix` / `dbtableprefix`.
- `lib/AppInfo/Application.php` — register service against `ActivityConnectionAdapter` so it queries the right DB.
- `lib/Settings/Admin.php` — inject and expose via initial state (`database_stats` key carries dedicated_connection + tables + retention_suggestion).
- `src/views/AdminSettings.vue` — new "Database" section with per-table size table, total row, contextual description, and the retention suggestion as an `NcNoteCard` with "Copy config snippet" action.


miaulalala:
@artonge can you check out this PR? I added Oracle support and some test coverage, in addition to Frank's changes,

PR #2550, opened 2026-05-02
docs: explain the optional dedicated activity database

karlitschek:
The Activity app already supports an `activity_db*`-prefixed connection that the AppInfo registers via `ConnectionFactory::createConnectionParams('activity_')`, falling back to the main Nextcloud DB when none of the keys are set. The behaviour is undocumented outside the source, which makes admins nervous about touching it on busy installations where the activity table grows fastest.

Add `docs/database.md` covering:

- when a dedicated connection is worth the operational cost,
- the full set of `activity_db*` config keys with types/defaults, including driver options and persistent-connection toggle,
- a worked `config.php` example for "same engine, separate host",
- a TLS driver-options snippet for MySQL,
- migration, backup and troubleshooting notes.

Link the new doc from the README under a new "Administration" section.


miaulalala (approved):
Also added the config options to the sample config: https://github.com/nextcloud/server/pull/59911

PR #2551, opened 2026-05-02
fix(security): escape link URLs in activity email templates

karlitschek:
The activity-digest and queued-mail templates interpolate the `link` attribute of rich-subject parameters and the home URL directly into the `href` of an `<a>` tag without HTML-escaping. Activity providers typically supply internal Nextcloud URLs, but defensive escaping closes a potential `"`-breakout XSS vector should a provider ever expose user-influenced link values.

Wrap the URLs in `htmlspecialchars(..., ENT_QUOTES, 'UTF-8')` in:
- DigestSender.php — "and N more" link, rich-subject link parameter
- MailQueueHandler.php — home link, rich-subject link parameter

Translation-string templates (`$l->t('...<a href="%s">...', ...)`) are left as-is: their URLs come exclusively from `linkToRouteAbsolute()`, and escaping inside the gettext source string would force translators to handle escaping for every locale.

### DIFF
--- a/lib/DigestSender.php
+++ b/lib/DigestSender.php
@@ -188,7 +188,7 @@ class DigestSender {
 			$andMoreText = $l10n->n('and %n more…', 'and %n more…', $skippedCount);
 			$url = $this->urlGenerator->linkToRouteAbsolute('activity.Activities.showList', [ 'filter' => 'all' ]);
 			$template->addBodyListItem(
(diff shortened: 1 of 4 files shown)


nickvergessen:
> ENT_QUOTES, 'UTF-8'

Is not needed in supported PHP versions.

PR #2562, opened 2026-05-06
fix(bulkReceive): honour admin email toggle and ISetting notification defaults

miaulalala:
## Summary

- **Email bug (main issue, stable33+master):** `bulkReceive()` was querying per-user email settings via `IUserConfig::getValuesByUsers()` — a raw DB query with no knowledge of the admin `enable_email` toggle. This caused activity emails to be sent to all users even when an admin had globally disabled them. Added the same `enable_email` guard that already exists in `UserSettings::getUserSetting()` and `filterUsersBySetting()`.

- **Notification bug:** When `canChangeNotification()` is `false`, `$userPushSettings` was left `null`, making `$notificationSetting` resolve to `null`. The guard `null !== false` is always true, so push notifications fired unconditionally. Fixed by deriving `$defaultPushEnabled` from `ISetting::isDefaultEnabledNotification()` and using it as the authoritative value when the user cannot change the setting.

## Test plan

- [x] `composer test:unit` passes (326 tests, 1580 assertions)
- [x] New test `testBulkReceiveNoMailWhenAdminEmailDisabled` verifies that `getValuesByUsers` is never called and no mail is queued when `enable_email=no`
- [x] Updated tests confirm corrected notification behaviour for non-user-configurable activity types
- [x] Backport to stable33 via GitHub

### DIFF
--- a/lib/Consumer.php
+++ b/lib/Consumer.php
@@ -16,6 +16,7 @@ use OCP\Activity\ISetting;
 use OCP\Config\IUserConfig;
 use OCP\Config\ValueType;
 use OCP\DB\Exception;
+use OCP\IAppConfig;
 
 class Consumer implements IConsumer, IBulkConsumer {
 
@@ -25,6 +26,7 @@ class Consumer implements IConsumer, IBulkConsumer {
 		protected UserSettings $userSettings,
 		protected NotificationGenerator $notificationGenerator,
 		protected IUserConfig $userConfig,
+		protected IAppConfig $appConfig,
 	) {
 	}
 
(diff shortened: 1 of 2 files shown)


lib/Consumer.php
```diff
@@ -78,13 +80,16 @@ public function bulkReceive(IEvent $event, array $affectedUserIds, ISetting $set
...
 
-		$userPushSettings = $userEmailSettings = $batchTimeSettings = null;
+		$userPushSettings = null;
 		if ($canChangePush === true) {
 			$userPushSettings = $this->userConfig->getValuesByUsers('activity', 'notify_notification_' . $event->getType(), ValueType::BOOL, $affectedUserIds);
 		}
 
-		if ($canChangeMail === true || $setting->isDefaultEnabledMail() === true) {
+		$userEmailSettings = $batchTimeSettings = null;
+		$emailEnabled = $this->config->getAppValue('activity', 'enable_email', 'yes') !== 'no';
```
come-nc: getAppValue is deprecated, you should inject IAppConfig (either from AppFramework to get it scoped or the global one if you need to access other apps config as well) and use the typed getters.

come-nc (dismissed):

miaulalala (commented):
Addressed: removed the unused `IConfig` dependency that was accidentally left in when `IAppConfig` was added. The constructor now only holds `IAppConfig` for the `enable_email` check, with typed `getValueString()`. All 41 Consumer tests pass.

lib/Controller/APIv2Controller.php
```diff
@@ -25,17 +25,31 @@
...
+	protected $objectType;
+
+	/** @var int */
+	protected $objectId;
+
+	/** @var string */
+	protected $user;
+
+	/** @var bool */
+	protected $loadPreviews;
```
skjnldsv: That look like a bad change ?

miaulalala: removed!

skjnldsv (approved):

PR #2563, opened 2026-05-06
fix(MailQueueHandler): check enable_email toggle before sending queued emails

miaulalala:
## Summary

The admin **Enable notification emails** toggle was only checked when *queuing* new emails (`Consumer::receive()` / `bulkReceive()`). The `MailQueueHandler::sendEmails()` background job that actually delivers emails had no knowledge of the toggle and would send all queued entries regardless.

This means: if emails were already sitting in `oc_activity_mq` when an admin disabled the toggle, they would still be sent on the next background job run.

The fix adds an early-return guard at the top of `sendEmails()`, consistent with the checks already in `UserSettings::getUserSetting()` and `filterUsersBySetting()`. The queue is intentionally left intact — if the admin re-enables emails, pending notifications can still be delivered.

Related to #2562 (which fixed the same missing check in `bulkReceive()`).

## Test plan

- [x] `composer test:unit` passes (327 tests)
- [x] New test `testSendEmailsSkipsWhenAdminEmailDisabled` verifies: return value is 0, mailer is never called, queue entries survive
- [x] Backport to stable33

### DIFF
--- a/lib/MailQueueHandler.php
+++ b/lib/MailQueueHandler.php
@@ -11,6 +11,7 @@ use OCP\Activity\IEvent;
 use OCP\Activity\IManager;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\Defaults;
+use OCP\IAppConfig;
 use OCP\IConfig;
 use OCP\IDateTimeFormatter;
 use OCP\IDBConnection;
@@ -51,6 +52,7 @@ class MailQueueHandler {
 		protected IFactory $lFactory,
 		protected IManager $activityManager,
 		protected IValidator $richObjectValidator,
+		protected IAppConfig $appConfig,
 		protected IConfig $config,
 		protected LoggerInterface $logger,
 		protected Data $data,
@@ -70,6 +72,10 @@ class MailQueueHandler {
 	 * @return int Number of users we sent an email to
 	 */
 	public function sendEmails(int $limit, int $sendTime, bool $forceSending = false, ?int $restrictEmails = null): int {
+		if (!$this->appConfig->getValueBool('activity', 'enable_email', true)) {
+			return 0;
+		}
+
 		// Get all users which should receive an email
 		$affectedUsers = $this->getAffectedUsers($limit, $sendTime, $forceSending, $restrictEmails);
 		if (empty($affectedUsers)) {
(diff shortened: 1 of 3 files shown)


lib/MailQueueHandler.php
```diff
@@ -70,6 +70,10 @@ public function __construct(
 	 * @return int Number of users we sent an email to
 	 */
 	public function sendEmails(int $limit, int $sendTime, bool $forceSending = false, ?int $restrictEmails = null): int {
+		if ($this->config->getAppValue('activity', 'enable_email', 'yes') === 'no') {
```
come-nc: Might as well inject IAppConfig and use that instead of the deprecated getAppValue, that will avoid having to adapt both the class and the test later on.

.github/workflows/npm-audit-fix.yml
```diff
@@ -67,22 +67,6 @@ jobs:
           npm ci
           npm run build --if-present
 
-      - name: Generate PR body
```
come-nc, line 71: Why is this file touched? If necessary it should be in a commit explaining why. Is that a pull from upstream (nextcloud/.github)?

come-nc (dismissed):
This is removing typing for no reason.
This looks like a git failure, like rebased on the wrong branch or something?

.github/workflows/npm-audit-fix.yml
miaulalala: ugh, I done goofed. Will revert those changes

lib/MailQueueHandler.php
```diff
@@ -70,6 +72,10 @@ public function __construct(
 	 * @return int Number of users we sent an email to
 	 */
 	public function sendEmails(int $limit, int $sendTime, bool $forceSending = false, ?int $restrictEmails = null): int {
+		if ($this->appConfig->getValueString('activity', 'enable_email', 'yes') === 'no') {
```
come-nc:
```suggestion
		if (!$this->appConfig->getValueBool('activity', 'enable_email', true)) {
```

lib/MailQueueHandler.php
miaulalala: Done — switched to `IAppConfig::getValueBool()` as suggested.

lib/MailQueueHandler.php
miaulalala: Applied, thanks!

come-nc (approved):

skjnldsv (approved):

PR #2568, opened 2026-05-06
refactor: clarify fileCreate() condition by making the specific case explicit

miaulalala:
## Summary

The original condition in `FilesHooks::fileCreate()` used a double-negative (De Morgan) form:

```php
// Before — hard to parse: "unless identifier is empty AND it's a public token"
if ($this->currentUser->getUserIdentifier() !== '' || !$this->currentUser->isPublicShareToken()) {
    // regular upload
} else {
    // public upload
}
```

Rewritten as a direct positive check, with the specific case (public upload) as the `if`-branch:

```php
// After — reads naturally
if ($this->currentUser->getUserIdentifier() === '' && $this->currentUser->isPublicShareToken()) {
    // public upload
} else {
    // regular upload
}
```

Also adds the missing fourth test case: a non-empty identifier combined with `isPublicShareToken=true` (a logged-in user uploading to a public share link) → treated as a regular upload.

No behaviour change.

## Test plan

- [x] All `testFileCreate` cases (4 data rows × 2 root tests = 6 assertions) — green

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -82,10 +82,10 @@ class FilesHooks {
 			return;
 		}
 
-		if ($this->currentUser->getUserIdentifier() !== '' || !$this->currentUser->isPublicShareToken()) {
-			$this->addNotificationsForFileAction($path, Files::TYPE_SHARE_CREATED, 'created_self', 'created_by');
-		} else {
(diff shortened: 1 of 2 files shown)


Antreesy (approved):
This I can comprehend 🦭

PR #2573, opened 2026-05-06
refactor(types): add native parameter/return types to Data::get and APIv2Controller

miaulalala:
## Summary
- `Data::get()`: type the 5 previously untyped parameters (`string $user`, `int $since`, `int $limit`, `string $sort`, `string $filter`) and add `array` return type; same treatment for the private `setOffsetFromSince()` helper
- `APIv2Controller`: replace eight `/** @var type */` docblock properties with native typed properties, each with a sensible default matching what `validateParameters()` would set

## Test plan
- [ ] `composer test:unit` passes (328 tests)
- Three test mocks that previously relied on `null` being assignable to untyped properties are updated to return a real string from `getUID()`

🤖 AI-Assisted-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/lib/Controller/APIv2Controller.php
+++ b/lib/Controller/APIv2Controller.php
@@ -60,15 +60,15 @@ class APIv2Controller extends OCSController {
 	 * @throws \OutOfBoundsException when no user is given
 	 */
 	protected function validateParameters(string $filter, int $since, int $limit, bool $previews, string $objectType, int $objectId, string $sort): void {
-		$this->filter = \is_string($filter) ? $filter : 'all';
+		$this->filter = $filter;
 		if ($this->filter !== $this->data->validateFilter($this->filter)) {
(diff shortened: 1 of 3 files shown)


ShGKme:
Why replace some `@var` with type annotations in the controller, but not all of them?

miaulalala:
> Why replace some `@var` with type annotations in the controller, but not all of them?

wdym?

ShGKme:
> wdym?

I meant, in `APIv2Controller` there are also methods with types only in `@param` docs comments.

ShGKme (approved):

miaulalala:
@copilot resolve the merge conflicts in this pull request

miaulalala:
@copilot fix the PHPUnit tests

miaulalala:
@copilot fix the static anlysis issues

PR #2590, opened 2026-05-12
chore(test): Migrate to @nc/e2e-test-server

artonge:
Signed-off-by: Louis Chmn <louis@chmn.me>

PR #2603, opened 2026-05-19
refactor(tests): migrate e2e test suite from Cypress to Playwright

miaulalala:
## Summary

- Replaces Cypress + `@nextcloud/cypress` with `@playwright/test` + `@nextcloud/e2e-test-server`
- All 9 original tests pass locally; the comment activity test remains skipped (pre-existing, pending a fix to the comments app)
- Adds 7 new tests covering the activity stream page and admin settings page
- Removes ~440 lines of Cypress config/boilerplate, adds ~750 lines of typed Playwright tests

## Motivation

Cypress CI has been flaky enough to block dependabot auto-merge. Playwright's architecture (no iframes, direct browser protocol) is more reliable, and `@nextcloud/e2e-test-server` already ships a Playwright integration that handles Docker container management and user creation.

## Key technical decisions

- **`start-server.mjs`**: spins up an isolated Docker container on port 8081; `reuseExistingServer` lets local runs skip the Docker step if the container is already up
- **`baseURL` trailing slash**: `index.php/` (with slash) is required for the login helper's Origin header construction
- **No leading slash for SPA routes**: navigating to `apps/activity/all` (without `/`) lets Playwright prepend the `index.php/` base, which the Vue Router requires to match its routes
- **`cssEscape()` helper**: `CSS.escape()` is a browser-only API; replaced with a Node-safe equivalent for attribute selector escaping
- **`hover()` before Actions click**: file rows only fully activate the Actions button on hover; without it the dropdown can silently fail to open
- **Disabled checkboxes**: email checkboxes are disabled when no SMTP is configured in the test instance; the settings test skips them
- **NcCheckboxRadioSwitch type="switch"**: wraps the input inside the label with no `for` attribute; must click the `.checkbox-radio-switch` wrapper instead of `label[for="..."]`

## Test plan

- [x] All sidebar activity tests pass (creation, favorite, share, rename, move, tag)
- [x] All settings persistence tests pass (form, notification frequency, activity summary)
- [x] Activity stream tests: entry rendering, heading/filter label, navigation filter, RSS feed toggle
- [x] Admin settings tests: email enable toggle visible, default settings grid visible, toggle persists across reloads
- [x] CI workflow updated (`playwright.yml` replaces `cypress.yml`)
- [ ] Close superseded PR #2590 (artonge's partial Cypress→e2e-test-server migration)


miaulalala:
Needs an update for the branch protection rules - the cypress summary is still required there and can go immediately after.

playwright.config.ts
```diff
@@ -0,0 +1,48 @@
...
+	},
+
+	projects: [
+		{
+			name: 'chromium',
+			use: { ...devices['Desktop Chrome'] },
+		},
+	],
+
+	webServer: {
```
susnux, line 38: I think you want something like this: https://github.com/nextcloud/server/blob/a525b3de367140a1a4c97425cf86bede6bb4ad74/playwright.config.ts#L27-L42

So it properly awaits the server to be ready (otherwise this gets flaky quite often)

playwright.config.ts
```diff
@@ -0,0 +1,48 @@
...
+// Note: createRandomUser() uses Docker exec into nextcloud-e2e-test-server_activity,
+// so it only works against the managed test container, not the main dev instance.
+const baseURL = process.env.PLAYWRIGHT_BASE_URL ?? 'http://localhost:8081'
+
+export default defineConfig({
+	testDir: './playwright',
+
+	fullyParallel: false,
+	forbidOnly: !!process.env.CI,
+	retries: process.env.CI ? 2 : 0,
```
susnux, line 20: This encourages flaky tests I would reduce this to 1

playwright.config.ts
```diff
@@ -0,0 +1,48 @@
...
+	workers: 1,
```
susnux, line 21: I would allow more locally to speed up local execution (but makes no sense on CI due to sharding):


```suggestion
	workers: process.env.CI ? 1 : undefined,,
```

playwright.config.ts
```diff
@@ -0,0 +1,48 @@
...
+
+	reporter: process.env.CI ? 'github' : 'html',
```
susnux: I would use multiple reporter for CI so you can download the reports and view them locally to check why something failed:


```suggestion
	reporter: process.env.CI ? [['blob'], ['dot'], ['github']] : 'html',
```

miaulalala:
Thanks @susnux! Switched to `wait: { stdout: /Nextcloud ready at/ }` to wait for the explicit ready log line from `start-server.mjs` instead of polling the URL. Also added `gracefulShutdown` while I was at it.

playwright.config.ts
```diff
@@ -0,0 +1,48 @@
...
+// Override with PLAYWRIGHT_BASE_URL to use an existing instance, e.g.:
+//   PLAYWRIGHT_BASE_URL=http://nextcloud.local npx playwright test
+// Note: createRandomUser() uses Docker exec into nextcloud-e2e-test-server_activity,
+// so it only works against the managed test container, not the main dev instance.
+const baseURL = process.env.PLAYWRIGHT_BASE_URL ?? 'http://localhost:8081'
+
+export default defineConfig({
+	testDir: './playwright',
+
+	fullyParallel: false,
```
susnux: Do tests fail with this enabled? I would encourage to use it as it leads to higher test quality

package.json
```diff
@@ -57,24 +54,18 @@
   },
   "devDependencies": {
     "@nextcloud/browserslist-config": "^3.1.2",
-    "@nextcloud/cypress": "^1.0.0-beta.15",
+    "@nextcloud/e2e-test-server": "^0.4.0",
     "@nextcloud/stylelint-config": "^3.2.1",
     "@nextcloud/vite-config": "^2.5.2",
-    "@testing-library/cypress": "^10.1.3",
-    "@types/dockerode": "^4.0.1",
+    "@playwright/test": "^1.49.0",
```
susnux: Why not latest version (1.60)?

playwright/start-server.mjs
susnux, line 1: Maybe you want to stop the server if aborted, in that case you can get some inspiration here:
https://github.com/nextcloud/server/blob/a525b3de367140a1a4c97425cf86bede6bb4ad74/tests/playwright/start-nextcloud-server.js#L62-L69

Also I did some optimization here to reduce flaky cases:
https://github.com/nextcloud/server/blob/a525b3de367140a1a4c97425cf86bede6bb4ad74/tests/playwright/start-nextcloud-server.js#L51-L59

maybe that helps

.github/workflows/playwright.yml
susnux, line 1: Here is an example how to do it with sharding and with merging reports afterwards so you can properly review whats failed locally:

https://github.com/nextcloud/server/blob/a525b3de367140a1a4c97425cf86bede6bb4ad74/.github/workflows/playwright.yml

playwright.config.ts
susnux: (I meant the "wait" part)

playwright.config.ts
miaulalala: Reduced to 1 in the latest commit.

playwright.config.ts
miaulalala: Kept at 1 unconditionally — even with SQLite WAL mode enabled, concurrent sessions for the same `admin` user cause Nextcloud-level 500s on login. The suite runs in ~1 minute so it's not a bottleneck for now. When it grows, the fix would be giving admin tests their own isolated user rather than sharing the built-in `admin` account.

playwright.config.ts
miaulalala: Done — switched to `[['blob'], ['dot'], ['github']]` in CI and added a `merge-reports` job to produce a downloadable HTML report after each run.

playwright.config.ts
miaulalala: Enabled. Tests don't fail with it — each test creates its own user. Kept `workers: 1` though since the single container can't handle parallel sessions.

package.json
miaulalala: Bumped to ^1.60.0.

playwright/start-server.mjs
miaulalala: Added SIGTERM/SIGINT handlers. In CI they call `stopNextcloud()`; locally they just exit cleanly without tearing down the container so the next run can reuse it.

.github/workflows/playwright.yml
miaulalala: Added the blob report + merge-reports pattern. Skipped sharding for now — with 17 tests and a single container the overhead isn't worth it yet, but the infrastructure is there if we grow the suite.

playwright.config.ts
miaulalala: For the admin users, yes

playwright.config.ts
susnux: Well you already have concurrent requests, a page load generates roughly 5 to 10 concurrent PHP requests:
- the page
- the styles (default, dark, light etc all loaded from PHP with session setup)
- some lazy loaded API

But you are probably right that its not the bottleneck :)

artonge (approved):

PR #2604, opened 2026-05-19
fix(NotificationGenerator): catch InvalidArgumentException from notification setters in prepare()

miaulalala:
## Summary

- `INotification` setters (`setRichSubject()`, `setParsedSubject()`) throw `\InvalidArgumentException` for empty/invalid values
- These previously escaped `prepare()`, triggering the NC39 deprecation: _"threw \InvalidArgumentException which is deprecated"_
- Fix catches the exception, logs it as an error, and throws `AlreadyProcessedException` to discard the unrenderable notification

## Test plan
- [x] Run unit tests: `composer test:unit`
- [x] Verify no `\InvalidArgumentException` escapes `prepare()` in the activity notifier logs

### DIFF
--- a/lib/NotificationGenerator.php
+++ b/lib/NotificationGenerator.php
@@ -118,7 +118,12 @@ class NotificationGenerator implements INotifier {
 		$event = $this->populateEvent($event, $languageCode);
 		$this->activityManager->setCurrentUserId(null);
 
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):
Reviewed the second commit. First one seems to be another PR

PR #2617, opened 2026-05-23
fix: fix null array key error in fixPathsForShareExceptions

icewind1991:
Starting with php 8.5 this is a hard error.

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -1047,7 +1047,9 @@ class FilesHooks {
 		$query = $queryBuilder->executeQuery();
 
 		while ($row = $query->fetch()) {
-			$affectedUsers[$row['share_with']] = $row['file_target'];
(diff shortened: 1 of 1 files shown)


lib/FilesHooks.php
```diff
@@ -1047,7 +1047,9 @@ protected function fixPathsForShareExceptions(array $affectedUsers, $shareId) {
 		$query = $queryBuilder->executeQuery();
 
 		while ($row = $query->fetch()) {
-			$affectedUsers[$row['share_with']] = $row['file_target'];
+			if ($row['share_with'] !== null) {
```
artonge, line 1050: When can it be null? Should we update the query instead to not fetch those?

icewind1991: Didn't look into things deeply enough to know

PR #2626, opened 2026-05-26
feat: adding an option to exclude users from the acitivity log

miaulalala:


miaulalala:
@Aiiaiiio can you add some tests and remove the "all" option for the exclusion please?

miaulalala (commented):
- We already have activity_expire_exclude_users (flat list of user IDs) and now `activity_log_exclude_users` (map of `user => [types]`). Same "exclude_users" naming, totally different shape. That's going to bite someone editing `config.php` by hand. Can we pick a less  confusable name or at least doc the shape clearly, either in docs and example config, or just the docs?
- `storeMail()`: you dropped the `$affectedUser `local var but still call `$event->getAffectedUser()` again for `amq_affecteduser`. Just keep the var, no need for the extra call.
- Non-array `$rule` silently returns "not excluded". Please add a debug log line for that case.

Otherwise the code and test coverage looks good!

Aiiaiiio:
@miaulalala point 3 is a good call, I fixed that. However

Point 1: Good point but I cannot find any such docs/sample config in this repo. I need a bit of help here.
Point 2: I think that variable is no longer needed. Before that was used in multiple places in this function, now just one. Declaring a variable is just noise. If you still think it would be better I'm happy to change that.

miaulalala (commented):
LGTM! Can't approve due to me being the PR author. @artonge could you approve please?

benjaminfrueh (approved):

miaulalala:
@Aiiaiiio will you handle the backports pls :pray:

PR #2627, opened 2026-05-28
fix(ActivityAppNavigation): use correct icon filter for NC34+ active nav items

miaulalala:
## Summary

- The `NcAppNavigationItem` component in `@nextcloud/vue` 9.8.0 introduced a new active state design for NC34+: active items now use `--color-primary-element-light` as background with `--color-main-text` for text, rather than the full primary colour with `--color-primary-element-text`
- The icon filter for active items was still using `--primary-invert-if-dark`, which is designed for icons on a full primary-coloured background — in dark mode on NC34+ this renders the icon dark/invisible
- Switch to `--background-invert-if-dark` for the non-legacy (NC34+) case, keeping `--primary-invert-if-dark` only for the `.app-navigation-entry--legacy` class applied on NC<34

Also added in https://github.com/nextcloud/activity/pull/2625 for NC34.

## Test plan

- [ ] Open the Activity app on NC34+ in dark mode — all navigation icons should be white/visible, including the active item
- [ ] Check the same on NC<34 (legacy) — active item icon should still use the primary-colour invert filter

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


AndyScherzinger (approved):

AndyScherzinger:
One thing we probably 8still) can't do (no judgment) is the active state icon switching given apps plug into the sidebar, so we face the same issue we have in files?

miaulalala:
> One thing we probably 8still) can't do (no judgment) is the active state icon switching given apps plug into the sidebar, so we face the same issue we have in files?

Unsure what you mean :grin: is the fix also needed in server?

AndyScherzinger:
> Unsure what you mean 😁

click on the side bar items in the contacts app and you see the icons changing

> is the fix also needed in server?

Yes, but I can't tell if the solution would be the same for both. Yet not in scope for this PR.

AndyScherzinger:
@miaulalala this is what I am referring to: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7273

miaulalala:
> @miaulalala this is what I am referring to: [nextcloud-libraries/nextcloud-vue#7273](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7273)

ah I see, let me check on monday what my fix actually looks like

miaulalala:
@AndyScherzinger checking if I can apply the "active" slot in a different PR.

miaulalala:
@AndyScherzinger would need a second icon variant, so I didn't implement that.

PR #2634, opened 2026-06-02
Improve activity feed design

jancborchardt:
Improve overall design of activity feed:
- Remove unnecessary h1 heading which duplicated the active nav entry. Keep it in the code for semantics/accessibility
- Make app-content the scroll container so the scrollbar isn't floating (this was just a bit unelegant)
- Make activity date headings sticky on scroll (with logic to look nice at the breakpoints)
- Add bottom whitespace below the end-of-feed message so it’s not stuck to the bottom

## Before

[Screencast From 2026-06-02 02-09-05.webm](https://github.com/user-attachments/assets/121c5b86-f171-450d-8ec0-7209b1514800)

## After

[Screencast From 2026-06-02 01-59-25.webm](https://github.com/user-attachments/assets/818b4c00-a7a9-4acc-9e4c-ad06c37c8fe9)

Breakpoint example:

[Screencast From 2026-06-02 02-01-05.webm](https://github.com/user-attachments/assets/48962435-3ad8-491d-8fa6-18941b1817e2)

AI-assisted: Claude Code (Opus 4.8)


jancborchardt:
@artonge @miaulalala any comments or feedback? :)

jancborchardt:
Rebased again :)

miaulalala:
> @artonge @miaulalala any comments or feedback? :)

I'll check it out and test it Monday. From the screencasts it looks like a nice refinement, thanks!

src/views/ActivityAppFeed.vue
```diff
@@ -4,7 +4,8 @@
 -->
 <template>
 	<NcAppContent class="activity-app">
-		<h1 class="activity-app__heading">
+		<!-- Kept for document semantics / screen readers, but visually hidden -->
+		<h1 class="activity-app__heading hidden-visually">
```
artonge, line 8: Let's not depend on a global class and have all the needed css here in this file :).

js/activity-app.mjs
artonge, line 1: No need to compile the assets in every commits.

src/views/ActivityAppFeed.vue
```diff
@@ -350,6 +352,8 @@ watch(props, () => {
 
 <style scoped lang="scss">
 .activity-app {
+	// Max width of the readable content column
+	--activity-feed-max-width: 924px;
```
artonge, line 356: Can you move the var declaration close its usage?
Also, why `924px`?

src/views/ActivityAppFeed.vue
```diff
@@ -366,15 +370,20 @@ watch(props, () => {
 	}
 
 	&__container {
+		// Full width so the scrollbar sits at the edge of app-content
+		height: 100%;
```
artonge: "Full width", but sets the `height`, is the comment correct and code?

src/views/ActivityAppFeed.vue
artonge: Ah, I see that 924 is from before 👍

src/components/ActivityGroup.vue
```diff
@@ -55,14 +57,44 @@ const fullDate = computed(() => {
 
 <style scoped lang="scss">
 .activity-group {
+	// Separate consecutive groups. Inside the <section> (not a margin between them)
+	// so the date stays pinned across the gap and the next date docks right as the
+	// group ends, instead of the push feeling early.
```
artonge: The comment is weirdly written, can you double-check it?

src/components/ActivityGroup.vue
```diff
@@ -55,14 +57,44 @@ const fullDate = computed(() => {
...
+		// width. It is the toggle clearance (--app-navigation-padding + clickable-area),
+		// less the space the content already has on its left: the centring gutter
+		// ((100cqi - column width) / 2, where 100cqi is the content-area width and the
+		// column width is --activity-feed-max-width) and its 12px padding. A little
+		// breathing room (+8px) minus that 12px padding nets to one negative grid
+		// baseline (-4px). Clamped to 0, so it never affects alignment with the entries
+		// once the gutter clears the toggle.
+		padding-inline-start: calc(max(
+			0px,
+			var(--app-navigation-padding, 8px) + var(--default-clickable-area)
```
artonge: Do we really need default values for the `var()`? I think we can assume that they are always available.

src/components/ActivityGroup.vue
```diff
@@ -55,14 +57,44 @@ const fullDate = computed(() => {
...
+				- var(--default-grid-baseline, 4px)
+				- max(0px, (100cqi - var(--activity-feed-max-width)) / 2)
+		));
```
artonge, line 85: Can't we come up with something simpler? I look quite complex for what we are trying to achieve.

src/views/ActivityAppFeed.vue
```diff
@@ -373,6 +373,11 @@ watch(props, () => {
 		text-align: center;
 	}
 
+	&__loading-indicator--end {
```
artonge: Is the new class necessary? Can't we add that to the existing class?

kra-mo (approved):
Cool, but we should actually standardize floating toolbars/headers as this stuff is a common headache.

artonge (changes requested):
Blocking until other comments are resolved :)

src/views/ActivityAppFeed.vue
jancborchardt: @artonge Using this class is our standard way of doing it, which we also use in other apps (like Mail etc.). If our method of hiding things visually ever changes, then we would need to change it everywhere we repeated the code.

artonge: Maybe cc @susnux, I don't have a strong opinion, but it feels strange to me, what do you think?

susnux: `hidden-visually` is a global class considered public API.
We use it everywhere for this purpose, see also https://docs.nextcloud.com/server/latest/developer_manual/html_css_design/css.html#css-classes

js/activity-app.mjs
jancborchardt: Ah right, wasn’t aware of the conventions. Removed the assets :)

src/views/ActivityAppFeed.vue
jancborchardt: Yeah, comment adjusted

src/components/ActivityGroup.vue
jancborchardt: Yeah, adjusted and shortened the wording.

src/components/ActivityGroup.vue
jancborchardt: Sounds good, some places we do use them, some places we don’t. Removed them now

src/components/ActivityGroup.vue
jancborchardt: This one I’m not sure about. It does try to do a lot, and I tried to come up with a more fitting comment to explain why it’s needed. If you test it in responsive mode and freely resize you can see it does a nice job at keeping things aligned.

src/views/ActivityAppFeed.vue
jancborchardt: Yeah, and it’s also weird to call it loading-indicator. Adjusted

jancborchardt:
@artonge all your points should be addressed, let me know if there’s anything else! :)

artonge (approved):

jancborchardt:
@miaulalala let me know if you have any feedback :)

jancborchardt:
@miaulalala rebased, in case you have any feedback before merging. :)

PR #2635, opened 2026-06-03
perf(mailqueue): cover `amq_latest_send` in the affecteduser index

solracsf:
## Summary

The email digest cron (`BackgroundJob\EmailNotification`, which runs on every cron tick) calls
`MailQueueHandler::getAffectedUsers()`:

```sql
SELECT amq_affecteduser, MIN(amq_latest_send)
  FROM activity_mq
  WHERE amq_latest_send < ?
  GROUP BY amq_affecteduser
  ORDER BY MIN(amq_latest_send) ASC
```

The amp_user index only contained `amq_affecteduser`, so `MIN(amq_latest_send)` could not be served from the index. MariaDB/MySQL therefore scanned the entire queue and built a temporary table to resolve the `GROUP BY` on every run..

This PR widens that index to (`amq_affecteduser`, `amq_latest_send`), which lets the engine resolve the aggregate with a loose index scan (Using index for group-by), reading one entry per user instead of the whole queue. amp_user is a strict prefix of the new index and becomes redundant, so it is dropped (no net change in index count).

**MariaDB 10.11**, `activity_mq` seeded with a ~60k-row backlog across 9 users:

| | index | `Handler_read_next` (index rows walked) | plan |
|---|---|---|---|
| **before** | `amp_user` | **60,108** | full index scan + `Using temporary; Using filesort` |
| **after** | `amp_user_send` | **0** (18 seeks) | `Using index for group-by` |

The result set is identical before and after (verified with `FORCE INDEX` vs `IGNORE INDEX`).

### DIFF
--- a/lib/Migration/Version8000Date20260603120000.php
+++ b/lib/Migration/Version8000Date20260603120000.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Activity\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Widen the `amq_affecteduser` index on `activity_mq` to also cover
+ * `amq_latest_send`.
+ *
+ * The email digest cron (`MailQueueHandler::getAffectedUsers()`) runs on every
+ * cron tick and does
+ *   SELECT amq_affecteduser, MIN(amq_latest_send)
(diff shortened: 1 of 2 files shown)


lib/Migration/Version8000Date20260603120000.php
```diff
@@ -0,0 +1,64 @@
...
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('activity_mq')) {
+			return null;
+		}
+
+		$table = $schema->getTable('activity_mq');
+		$changed = false;
```
miaulalala: nit: you don't need to handle this, you can just return the `$schema` iirc.

miaulalala (approved):

lib/Migration/Version8000Date20260603120000.php
solracsf: Fixed

miaulalala:
backport to stable32

PR #2784, opened 2026-07-09
fix: do not fail moves when the old location cannot be resolved

solracsf:
Related: https://github.com/nextcloud/groupfolders/issues/4548
This has been found by analyzing the stack traces in that issue.

When the pre-rename hook cannot resolve the old location of a move by file id (NotFoundException from `View::getPath()`, e.g. on instances with inconsistent mount or file cache state), `fileMove()` aborted after setting `$this->moveCase` but before collecting `$this->oldAccessList`.

The post-rename hook then crashed with "`TypeError: array_keys(): Argument #1 ($array) must be of type array, null given`" in `fileMoving()`. Since `OC_Hook::emit()` only catches Exceptions, the TypeError failed the whole WebDAV MOVE request after the files had already been moved, leaving clients out of sync with the server.

Collect the state in `fileMove()` under a NotFoundException guard that disables the post hook handling instead, and bail out of `fileMoving()` if the old access list was never collected.

Actually, this results in things like files being moved correctly on disk, but not on `oc_filecache`, and only a manual scan will fix it.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -43,7 +43,7 @@ class FilesHooks {
 
 	/** @var string|bool */
 	protected $moveCase = false;
-	/** @var array */
+	/** @var array|null */
 	protected $oldAccessList;
 	/** @var string */
 	protected $oldParentPath;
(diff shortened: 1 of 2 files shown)


miaulalala (approved):
Makes sense

PR #2813, opened 2026-07-22
fix(files): do not publish incompletely built activity events

karlitschek:
When building the activity event threw \InvalidArgumentException, the error was logged but execution fell through, so the partially initialized event was still stored in the stream, queued for email and pushed as a notification. Return after logging instead.

## 🤖 AI (if applicable)


karlitschek:
@nickvergessen

CarlSchwan (approved):

PR #2814, opened 2026-07-22
fix(mail): always reset current user after sending activity emails

karlitschek:
MailQueueHandler::sendEmailToUser() and DigestSender::sendDigestForUser() set the current user on the activity manager but only reset it on the success path. Early returns (no activities for the digest) and failures (mailer errors, exceptions while building the mail — which the digest loop catches and continues) left the activity manager impersonating that user for the remaining users of the same cron run, leaking activity visibility between users.

Wrap the impersonated sections in try/finally so the current user is always reset.

## 🤖 AI (if applicable)


karlitschek:
@nickvergessen

miaulalala (approved):

PR #2816, opened 2026-07-22
perf(api): avoid up to 200 notification queries when viewing file act…

karlitschek:
…ivities

When viewing the activities of a file, every returned activity was turned into a markProcessed() call, each of which queries and deletes against the notification storage — up to 200 round trips per sidebar request, mostly for notifications that do not exist.

Skip activities of one's own actions (they never generate a notification) and bail out with a single COUNT query when the user has no notifications at all, which is the common case.

## 🤖 AI (if applicable)


karlitschek:
@nickvergessen

PR #2869, opened 2026-08-18
fix(navigation): fill the active filter icon, outline the rest

### LINKED ISSUE #2662: Use filled icons in the part of Activity

pringelmann:
Resolves #2662

The nav icons were a mix of outlined and filled, with nothing tying that to which filter you're on. Each filter id now has both variants from `@mdi`, and the icon slot picks one off the `NcAppNavigationItem` `active` prop:

| Before | After |
| --- | --- |
|image|image|

The map has to be here rather than in the backend: `getIcon()` returns one URL, and a second state would mean changing public OCP API. Filters not in the map keep their backend URL and stay put when selected. One example is Deck, where we decided to keep the actual app icon as a real solid mark, plus Tasks and anything else third-party.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI (tests)


src/views/ActivityAppNavigation.vue
```diff
@@ -148,6 +192,7 @@ async function copyRSSLink() {
 	}
 
 	.app-navigation-entry {
+
```
pringelmann, line 208: auto-format on save

artonge (approved):

AndyScherzinger (approved):

jancborchardt (commented):
Nice, much better :)

PR #2880, opened 2026-08-25
fix: ignore playwright dev files in release archive

skjnldsv:
Update of the nextcloudignore to avoid shipping dev files


skjnldsv:
I see this is a new file here.
In the long run, the release script won't clean anything for you. It will assume Dev's maintenance of the Nextcloudignore file accordingly.

If there are other files you think should be excluded, please adjust this pr.

Thanks 😊

artonge (approved):

